PR #4, opened 2019-07-29
Refactor pegen.py into multiple files

pablogsal:
To organize the project, split pegen.py into individual components
and add a package structure with a main entrypoint.

Note
-----

This PR may feel a bit invasive as is setting an initial structure to the project that you may not 100% like. If you only require some changes, pease, advise what structure would you prefer and we can iterate here together. If you think you prefer to do this yourself because more changes are required, feel free to close the PR :)


pablogsal:
Hummm, it seems that I need to rebase to pick up the latest changes. Before doing that I would like to know if you like the direction of the PR, as we can just close it if you prefer to do the refactor yourself.

Makefile
```diff
@@ -18,8 +18,8 @@ parse.o: parse.c pegen.h v38tokenizer.h
 pegen.o: pegen.c pegen.h
 	$(CC) $(CFLAGS) -c $(INCLUDES) pegen.c
 
-parse.c: $(GRAMMAR) pegen.py
-	$(PYTHON) pegen.py -q -c $(GRAMMAR) -o parse.c
+parse.c: $(GRAMMAR) pegen
```
gvanrossum: Does that work? Shouldn't the dependency be something like `pegen/*.py`?

pablogsal: Yeah, this file was not added correctly to the commit, apologies. I will update the Makefile to also move all the generated files and compilation to `pegen/`

pablogsal:
I have rebased over the latest version and

* [x] Moved headers and source files to pegen/
* [x] Updated Makefile to generate object files and shared objects in pegen/
* [x] Updated Makefile targets
* [x] Update .gitignore
* [x] Moved `simple_parser_main` to `pegen/parser.py`

By the way, to make sure nothing is merged without passing travis, you probably want to activate branch protection in the repo and check the travis checks as "required". Here are the instructions:

https://help.github.com/en/enterprise/2.15/admin/developer-workflow/configuring-protected-branches-and-required-status-checks#enabling-required-status-checks

pablogsal:
> OK, LG. I have some follow-up changes to `Makefile` (the deps and imports in the `test`, `dump` and `time` targets are wrong) but I'll do those as a follow-up

Thanks @gvanrossum! I could not test them correctly this because I got a lot of undefined symbols and I did not have time to investigate until tomorrow (maybe the Makefile depends too much on clang/MacOS? and I am using gcc on Linux).

gvanrossum:
@pablogsal The Makefile is definitely a Mac hack. I think given the stage the project is in that's fine (though if you find a fix for Linux that doesn't break Mac it is of course welcome!). I expect to rewrite most of the code as part of the blogging project.

PR #7, opened 2019-07-31
Refactor grammar to start using visitor pattern in parser generators

pablogsal:
Work in progress (draft PR) for the refactor work to use the visitor pattern. I have opened a draft PR for early feedback and iteration if needed. I opened the draft also to check if the travis checks are fixed or not. Feel free to ignore this PR until is ready if you want.

- [x] Create temporary files (like grammar2) start the refactoring without breaking old tests (to compare old and new parsers).
- [x] Eliminate parser-related methods for grammar node classes.
- [x] Add a way to iterate over grammar node children (for default visit).
- [x] Add a basic grammar visitor to inherit from.
- [x] Create a Python parser generator using a subclass from the grammar visitor using the methods that were in the grammar nodes before.
- [x] Create a C parser generator using a subclass from the grammar visitor using the methods that were in the grammar nodes before.
- [x] Substitute old generators and grammar files for the new ones.

- [x] Test for the default grammar visitor covering simple grammar nodes with multiple options.
- [ ] Test for the Python generator using visitor pattern?.

Extras
---------
- [x] Add grammar tree visualizer to help debugging

Notes
--------

For iterating over grammar nodes I am treating the nodes as iterators but maybe we should have them expose a `children` attribute or similar.

Example of the grammar visualizer:
```
start: 'a'| ('b' ['c'])
```
```
python -m pegen.grammar_visualizer example.gram
 Rules┐
│     └Rule(start)┐
│                 │   ┌Alt┐
│                 │   │   └NamedItem┐
│                 │   │             └StringLeaf('a')
│                 └Rhs┤
│                     └Alt┐
│                         └NamedItem┐
│                                   └Group┐
│                                         └Rhs┐
│                                             │   ┌NamedItem┐
│                                             │   │         └StringLeaf('b')
│                                             └Alt┤
│                                                 └NamedItem┐
│                                                           └Opt┐
│                                                               └Rhs┐
│                                                                   └Alt┐
│                                                                       └NamedItem┐
│                                                                                 └StringLeaf('c')
```


pablogsal:
By the way, note that we only have 1 Travis Check (yay!). I had to deactivate the `travis.org` one istead as the `travis.com`  is the one you marked as "required".

test/test_pegen.py
```diff
@@ -367,3 +368,108 @@ def test_cut():
     assert node == [TokenInfo(OP, string='(', start=(1, 0), end=(1, 1), line='(1)'),
                     [TokenInfo(NUMBER, string='1', start=(1, 1), end=(1, 2), line='(1)')],
                     TokenInfo(OP, string=')', start=(1, 2), end=(1, 3), line='(1)')]
+
+
+class TestGrammarVisitor:
```
gvanrossum: Why not continue the pytest style test functions, without the test case class?

test/test_pegen.py
```diff
@@ -367,3 +368,108 @@ def test_cut():
...
+    class Visitor(pegen.grammar2.GrammarVisitor):
+        def __init__(self):
+            self.n_nodes = 0
+
+        def visit(self, node):
+            self.n_nodes += 1
+            super().visit(node)
+
+    def test_parse_trivial_grammar(self):
+        # GIVEN
```
gvanrossum: Is this GIVEN / WHEN / THEN from your corporate style guide? I find it rather verbose.

pegen/grammar_visualizer.py
```diff
@@ -0,0 +1,92 @@
...
+
+def main() -> None:
+    args = argparser.parse_args()
+    with open(args.filename) as file:
+        tokenizer = Tokenizer(grammar_tokenizer(tokenize.generate_tokens(file.readline)))
+        parser = GrammarParser(tokenizer)
+        rules = parser.start()
+        visitor = GrammarPrinter()
+        visitor.print_grammar(rules)
+
```
gvanrossum: You need two blank lines here as well.

pegen/grammar_visualizer.py
```diff
@@ -0,0 +1,92 @@
...
+            self.print_grammar(child, next_indent, next_last)
+
+
+def main() -> None:
+    args = argparser.parse_args()
+    with open(args.filename) as file:
+        tokenizer = Tokenizer(grammar_tokenizer(tokenize.generate_tokens(file.readline)))
+        parser = GrammarParser(tokenizer)
+        rules = parser.start()
+        visitor = GrammarPrinter()
```
gvanrossum: For this line and the next you don't need to be inside the with-statement.

pegen/grammar_visualizer.py
```diff
@@ -0,0 +1,92 @@
+import argparse
```
gvanrossum: This is a nice tool. A nomenclature nit: It doesn't visualize the grammar but the AST (for the grammar). And that's useful separately.

pegen/grammar_visualizer.py
```diff
@@ -0,0 +1,92 @@
...
+import tokenize
+import enum
+
+from pegen.tokenizer import Tokenizer
+from pegen.tokenizer import grammar_tokenizer
+from pegen.grammar2 import GrammarParser
+
+
+argparser = argparse.ArgumentParser(prog='pegen', description="Experimental PEG-like parser generator")
```
gvanrossum: description=what??

pegen/grammar_visualizer.py
```diff
@@ -0,0 +1,92 @@
...
+
+class GrammarPrinter:
+
+    def children(self, node):
+        for value in node:
+            if isinstance(value, list):
+                yield from value
+            else:
+                yield value
+    def nb_children(self, node):
```
gvanrossum: Blank line missing.

pegen/grammar_visualizer.py
```diff
@@ -0,0 +1,92 @@
...
+    def name(self, node):
+        if (name := getattr(node, "name", None)):
+            return node.__class__.__name__ + f"({node.name})"
+        elif (value := getattr(node, "value", None)):
+            return node.__class__.__name__ + f"({node.value})"
+        else:
+            return node.__class__.__name__
+
+
+    def print_grammar(self, current_node, indent='', last=Direction.NEUTRAL, printer=print):
```
gvanrossum: This method has too much whitespace in it (to my taste :-).

Do you have a reference for the algorithm?

pegen/grammar_visualizer.py
```diff
@@ -0,0 +1,92 @@
...
+        while up and sum(child_sizes[node] for node in down) < sum(child_sizes[node] for node in up):
+            down.append(up.pop())
+
+        for child in up:
+            next_last = Direction.UP if up.index(child) == 0 else Direction.NONE
+            next_indent = (f"{indent}{' ' if last == Direction.UP else '│'}"
+                           f"{' ' * len(self.name(current_node))}")
+            self.print_grammar(child, next_indent, next_last)
+
+        if last == Direction.UP:
```
gvanrossum: These two if/elif blocks look low in density. Maybe the first can use a pre-populated dict? The second could be an if-expression.

pegen/grammar2.py
```diff
@@ -0,0 +1,464 @@
...
+
+from pegen.parser import memoize, Parser
+from pegen.tokenizer import exact_token_types
+
+
+class GrammarNode:
+    pass
+
+
+class GrammarVisitor(object):
```
gvanrossum: No need to inherit from object.

pegen/grammar2.py
```diff
@@ -0,0 +1,464 @@
...
+
+    def visit(self, node):
+        """Visit a node."""
+        method = 'visit_' + node.__class__.__name__
+        visitor = getattr(self, method, self.generic_visit)
+        return visitor(node)
+
+    def generic_visit(self, node):
+        """Called if no explicit visitor function exists for a node."""
+        for value in node:
```
gvanrossum: Huh? I don't think all nodes are iterable?

pablogsal:
Example of new improved grammar AST printer:

```
│     ┌Rule(start)──Rhs──Alt──NamedItem──NameLeaf(a)
│     │
 Rules┤
│     │                 ┌NamedItem──NameLeaf(a)
│     │                 │
│     └Rule(a)──Rhs──Alt┤
│                       ├NamedItem──StringLeaf('b')
│                       │
│                       └NamedItem──StringLeaf('+')
│
```

pegen/grammar_visualizer.py
pablogsal: I have done a refactor of everything, hopefully, is now a bit more clear. If not I can explain or add more comments.

test/test_pegen.py
pablogsal: I can remove the class if you prefer. I normally use classes (still pytest-style classes with no extra inheritance or anything) to group related tests together in the same file, so is a bit more organized and then I can select to collapse the things that I am not looking at.

pegen/grammar2.py
pablogsal: This is just a temporary hack before we set up how the nodes expose their children. I am implementing `__iter__` on all grammar nodes to obtain all subnodes. We can change this to be a descriptor or a method (node.children or similar) later.

pegen/grammar_visualizer.py
```diff
@@ -1,82 +1,114 @@
...
 
+    @functools.lru_cache
+    def _total_number_of_children(self, node):
+        """Get the total number of children nodes under a given one recursively"""
+        return sum(self._total_number_of_children(child) for child in self.children(node)) + 1
 
-    def print_grammar(self, current_node, indent='', last=Direction.NEUTRAL, printer=print):
+    def _segment_children(self, node):
+        """Separate children of a given node in two groups so every group has roughly
+        the same number of nodes under it."""
```
gvanrossum: I prefer multi-line docstrings having a blank line after the first (so the first line is the title or subject, similar to commit messages, where the first line is special (e.g. `git log --oneline`).

pegen/grammar_visualizer.py
```diff
@@ -1,82 +1,114 @@
...
+
+class ASTGrammarPrinter:
+    _direction_to_start_shape = {
+        Direction.UP: '┌',
+        Direction.DOWN: '└',
+        Direction.NEUTRAL: ' ',
+        Direction.NONE: '├'
+    }
+
+    @staticmethod
```
gvanrossum: Just because something doesn't use self, why make it a static method? That decorator is meant for things you can call directly on the class, but here `GrammarPrinter.children(node)` doesn't make sense.

pegen/grammar_visualizer.py
```diff
@@ -1,82 +1,114 @@
...
-    def print_grammar(self, current_node, indent='', last=Direction.NEUTRAL, printer=print):
+    def _segment_children(self, node):
+        """Separate children of a given node in two groups so every group has roughly
+        the same number of nodes under it."""
 
-        child_sizes = {child: self.nb_children(child) for child in self.children(current_node)}
+        child_sizes = {child: self._total_number_of_children(child)
+                       for child in self.children(node)}
 
         up = sorted(child_sizes, key=lambda key: child_sizes[key])
```
gvanrossum: Why sort? I'd rather see the children displayed in the order in which they are stored in their parent. That order is significant in PEG grammars (since earlier alternatives must be tried before later ones).

pegen/grammar_visualizer.py
```diff
@@ -1,82 +1,114 @@
...
+        # TODO: Make the grammar nodes be able to return this instead of hardcoding attributes
         if (name := getattr(node, "name", None)):
             return node.__class__.__name__ + f"({node.name})"
         elif (value := getattr(node, "value", None)):
             return node.__class__.__name__ + f"({node.value})"
         else:
             return node.__class__.__name__
 
+    @functools.lru_cache
+    def _total_number_of_children(self, node):
```
gvanrossum: Why bother giving certain methods an underscore? It's all implementation anyway, nothing here is public API except `print_grammar_ast()`, and the whole thing is part of a one-off script, not a library.

pegen/grammar_visualizer.py
```diff
@@ -1,82 +1,114 @@
...
+        """Visit a group of children recursively printing them accordingly taking into account if they
+           must be printed over the parent node or below it (indicated by *direction*)."""
+
+        for child in group:
+            next_last = direction if child is boundary else Direction.NONE
+            next_indent = self._next_indent(indent, n_spaces, direction, last_direction)
+            self._print_nodes_recursively(child, next_indent, next_last, printer)
 
-        printer(f'{indent}{start_shape}{self.name(current_node)}{end_shape}')
+        # Add an extra line between nodes when we branch out so they don't appear super packed.
```
gvanrossum: There is an anomaly caused by this part. Look at this output:
```
│     │                  ┌Alt──NamedItem──NameLeaf(list)
│     │                  │
│     │                  ├Alt──NamedItem──NameLeaf(tuple)
│     │                  │
│     │                  ├Alt──NamedItem──NameLeaf(group)
│     │                  │
│     ├Rule(primary)──Rhs┤
│     │                  ├Alt──NamedItem──NameLeaf(NUMBER)
│     │                  │
│     │                  ├Alt──NamedItem──NameLeaf(STRING)
│     │                  │
│     │                  └Alt──NamedItem──NameLeaf(NAME)
```
Note that there's an extra "blank" line before the `Rule(primary)----Rhs` in the middle that shouldn't be there (at least, it breaks the neat visual array of children).

pablogsal:
I am playing a lot with the visualizer as well when I can make some time to work on this :)

 If we renounce to have centered parents, we can simplify the code (check commit 5208a418110624428f02bef) substantially and have this kind of output:

```
└──Rules
   ├──Rule(start)
   │  └──Rhs
   │     └──Alt
   │        └──NamedItem
   │           └──NameLeaf(a)
   └──Rule(a)
      └──Rhs
         └──Alt
            ├──NamedItem
            │  └──Repeat1
            │     └──NameLeaf(a)
            └──NamedItem
               └──StringLeaf('b')
```

pablogsal:
Commit 06ec369 adds a first working draft version of the python generator using several visitors. Please, check out the general architecture and tell me what you think. Do you have some other structure in mind?

pegen/parser_generator.py
```diff
@@ -163,6 +286,59 @@ def name_loop(self, node: Plain, is_repeat1: bool) -> str:
...
+    origname = name
+    counter = 0
+    while name in names:
+        counter += 1
+        name = f"{origname}_{counter}"
+    names.append(name)
+    return name
+
+
 def compute_nullables(rules: Dict[str, Rule]) -> None:
```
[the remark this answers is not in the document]
pablogsal: The reason this was not working is because instead of making the parser return a dictionary I am returning a new "Rules" node that have a dictionary inside (the reason is in this way the parser always return a tree of nodes instead of a collection of trees (the rule dictionary)).

This makes things slightly more homogeneous but I can revert that part and just work everywhere with the dictionary. Check 7c8aad2 for the fix, for example.

pegen/parser_generator.py
```diff
@@ -8,6 +8,7 @@
 from pegen.grammar import Rhs
 from pegen.grammar import Alt
 from pegen.grammar import NamedItem
+from pegen.grammar import GrammarVisitor
```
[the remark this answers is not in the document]
pablogsal: The reason is to reduce future diff if you add more imports, but I am ok with placing them on the same line :)

pablogsal:
Nice, I liked it more on this last iteration after addressing your feedback. Especially getting rid of the partial visitors and unifying them makes it more clear.

pegen/__init__.py
```diff
@@ -10,7 +11,6 @@
 __all__ = [
```
gvanrossum: Maybe we should make `__init__.py` totally empty? The way it is now you can't run a parser without importing all the generator code.

pegen/python_generator.py
```diff
@@ -0,0 +1,188 @@
...
+"""
+MODULE_SUFFIX = """
+
+if __name__ == '__main__':
+    from pegen import simple_parser_main
+    simple_parser_main(GeneratedParser)
+"""
+
+
+class PythonElementVisitor(GrammarVisitor):
```
gvanrossum: Seems dead code?

pegen/python_generator.py
```diff
@@ -0,0 +1,188 @@
...
+    def visit_Group(self, node):
+        return self.visit(node.rhs)
+
+
+class PythonParserGenerator(ParserGenerator, GrammarVisitor):
+    def __init__(self, rules: grammar.Rules, file: Optional[IO[Text]]):
+        super().__init__(rules.rules, file)
+        self.callmakervisitor = PythonCallMakerVisitor(self)
+
+    def generate_python_module(self, filename: str) -> None:
```
gvanrossum: Does this still need "Python" in its name?

pablogsal:
@gvanrossum I have implemented the C parser generator using the visitor pattern and address previous feedback and done some minor cleanups. I also checked in a mac laptop that the makefile works as intended. 

There is still some state in the grammar file (`collect_vars`, `collect_todo`...etc) that maybe we can move out but I suggest doing so in another PR because this one is getting very big already.

I have squashed all the commits into a final one to make the rebase a bit easier because it was causing too many steps that needed to be resolved manually.

gvanrossum:
(Huh, PR branches should be auto-deleted after merging?)

pablogsal:
> (Huh, PR branches should be auto-deleted after merging?)

I deleted it manually (the upstream branch in my fork). In this way, I track what branches I can also delete locally. Usually, this does not happen automatically.

gvanrossum:
Oh wait, you deleted the branch in your fork. I guess the PR system has no say over that. The setting I found ("Automatically delete head branches") only applies to PR branches in the same repo.

pablogsal:
> The setting I found ("Automatically delete head branches") only applies to PR branches in the same repo.

Oh, I see, I did not know about that option. This sounds useful if we open a lot of PRs from this repository itself so there are not there laying around. In my previous comment I thought you were referring to this message:

>  pablogsal deleted the pablogsal:visitor_Refactor  branch 2 hours ago

gvanrossum:
I was referring to that, but I hadn’t read it carefully...

PR #8, opened 2019-08-02
Finish making pegen/ mypy-clean (and a few more cleanups)

gvanrossum:
This was done in a bit of haste, so there may be some issues.

@pablogsal I'm not merging this so as not to cause you more merge grief with #7.

I also cleaned up a few things, e.g. `from pegen import` should be `from pegen.parser.import` in a few cases. (Also `__init__.py` should be empty, but I didn't do that.)

**NOTE:** This requires a non-master version of mypy, https://github.com/python/mypy/pull/6899. Because of a bug there, whenever there's code of the form
```py
if foo := self.foo():
    some_call(foo)
```
where `self.foo()` returns `Optional[X]` but `some_call()` requires plain `X`, I had to add
```py
    assert foo
```
before the call. These should be taken out once that bug is fixed. (Ideally I should have marked them with `# XXX` or something, but I thought of that too late.)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,5 @@ pegen/parse
 pegen/parse.c
 
 **/__pycache__
+
+.dmypy.json
--- a/mypy.ini
+++ b/mypy.ini
@@ -1,0 +1,4 @@
+[mypy]
+files = pegen, story*
+follow_imports = error
+disallow_any_generics = true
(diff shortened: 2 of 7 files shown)


gvanrossum:
Eh, I'm just going to land this.

gvanrossum:
(Landed directly from my local repo, so closing.)

PR #9, opened 2019-08-05
Add a --compile-extension command to build generated C extension

pablogsal:
To allow building the generated extesion when using the --cpython
command, add a --compile-extension command that automatically generates
an extension module (shared object). In this way, we can also build the
extensions easily on other platforms (like Linux).

### DIFF
--- a/pegen/__main__.py
+++ b/pegen/__main__.py
@@ -8,11 +8,17 @@ Search the web for PEG Parsers for reference.
 from __future__ import annotations  # Requires Python 3.7 or later
 
 import argparse
+import pathlib
+import shutil
 import sys
 import time
 import token
 import tokenize
 import traceback
+
+from distutils.core import Distribution, Extension
+from distutils.command.clean import clean
+from distutils.command.build_ext import build_ext
 
 from typing import Final
 
@@ -23,6 +29,7 @@ from pegen.tokenizer import Tokenizer
 from pegen.tokenizer import grammar_tokenizer
 from pegen.grammar import GrammarParser
 
+MOD_DIR = pathlib.Path(__file__)
 
 def print_memstats() -> bool:
     MiB: Final = 2 ** 20
@@ -58,6 +65,7 @@ argparser.add_argument('-q', '--quiet', action='store_true', help="Don't print t
 argparser.add_argument('-v', '--verbose', action='count', default=0,
                        help="Print timing stats; repeat for more debug output")
 argparser.add_argument('-c', '--cpython', action='store_true', help="Generate C code for inclusion into CPython")
+argparser.add_argument('--compile-extension', action='store_true', help="Compile generated C code into an extension module")
 argparser.add_argument('-o', '--output', metavar='OUT',
                        help="Where to write the generated parser (default parse.py)")
 argparser.add_argument('filename', help="Grammar description")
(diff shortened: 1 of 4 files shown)


pegen/__main__.py
```diff
@@ -8,12 +8,18 @@
 from __future__ import annotations  # Requires Python 3.7 or later
 
 import argparse
+from pathlib import Path
```
gvanrossum: Just use `import pathlib` and write `pathlib.Path` twice.

gvanrossum:
And trying to add a `-o` flag just made it crash elsewhere:
```
$ python3 -m pegen -c --compile-extension data/cexpr.gram -o lala.c
Clean Grammar:
  start[mod_ty]: a=(stmt)* $ { Module ( a , NULL , p -> arena ) }
  stmt[stmt_ty]: a=expr_stmt { a }
  expr_stmt[stmt_ty]: a=expr NEWLINE { _Py_Expr ( a , EXTRA ( a , a ) ) }
  expr[expr_ty]: (l=expr '+' r=term { _Py_BinOp ( l , Add , r , EXTRA ( l , r ) ) } | l=expr '-' r=term { _Py_BinOp ( l , Sub , r , EXTRA ( l , r ) ) } | t=term { t })
  term[expr_ty]: (l=term '*' r=factor { _Py_BinOp ( l , Mult , r , EXTRA ( l , r ) ) } | l=term '/' r=factor { _Py_BinOp ( l , Div , r , EXTRA ( l , r ) ) } | f=factor { f })
  factor[expr_ty]: ('(' e=expr ')' { e } | a=atom { a })
  atom[expr_ty]: (n=NAME { n } | n=NUMBER { n } | s=STRING { s })
/Users/guido/pegen/pegen/pegen.c:6:1: warning: unused function 'token_name' [-Wunused-function]
token_name(int type)
^
1 warning generated.
  File "/usr/local/lib/python3.8/runpy.py", line 192, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "/usr/local/lib/python3.8/runpy.py", line 85, in _run_code
    exec(code, run_globals)
  File "/Users/guido/pegen/pegen/__main__.py", line 174, in <module>
    main()
  File "/Users/guido/pegen/pegen/__main__.py", line 127, in main
    shutil.move(cmd.get_ext_fullpath(extension_name),
  File "/usr/local/lib/python3.8/shutil.py", line 739, in move
    raise Error("Destination path '%s' already exists" % real_dst)
shutil.Error: Destination path './lala.cpython-38-darwin.so' already exists
$
```

pablogsal:
> TypeError: expected str, bytes or os.PathLike object, not NoneType

Fixed this on the latest ammend.

> shutil.Error: Destination path './lala.cpython-38-darwin.so' already exists

This error was due to the fact that the file already existed (the extension was already generated) and was going to be overwritten. In the latest ammend I have changed the code to always overwrite.

pablogsal:
I have tested the latest version on a mac system and it works now (both with -o and without). In addition, I have also:

1. Use the new build code in the Makefile.
2. Added a simple test to the test suite to check the c parser over some test cases.
3. Some minor refactors and cleanups.

Regarding the makefile, I have left your hardcoded path for Python:

```
TOP = /usr/local
PYTHON = $(TOP)/bin/python3.8
```

but in the future we can use just "python" and invoke it from a 3.8 virtualenv or something (which is what I am doing in Linux).

gvanrossum:
Very cool. Can we add this to Travis-CI too? Also just noticed that this broke the mypy build -- can you fix that?

pablogsal:
> Can we add this to Travis-CI too?
We have a test testing indirectly the `compile_c_extension` function added in this PR (the one testing the C parser). Do you have in mind something else to add to Travis-CI? Maybe separate the Python and the C tests?

> Also just noticed that this broke the mypy build -- can you fix that?

Yep, will prepare a PR soon :)

PR #10, opened 2019-08-05
Fix grammar visualizer after latest changes

pablogsal:
Fix grammar visualizer after some of the latest refactors and add better
error messages if the grammar file fails to be parsed.


pablogsal:
I have added some basic tests for the grammar visualizer. More to come soon in another PR!

PR #11, opened 2019-08-05
Run mypy in the CI build

pablogsal:
Note
----

Mypy does still not support PEP 572 in the master branch, but to avoid
pushing code that breaks mypy check, we can still run mypy in the test
suite by installing the PR branch where PEP 572 support is being added.

Once it lands, we can switch to install the master branch or the latest
version once is released.


pablogsal:
> Please remove the -v flag, it is too verbose. :-)

Done! :)

PR #12, opened 2019-08-05
Ignore the types from distutils.command until they are added to typeshed

pablogsal:
Currently the mypy build is failing because the typeshed information for
distutils.command is empty:

https://github.com/python/typeshed/blob/master/stdlib/2and3/distutils/command/clean.pyi

Before this is address upstream, type ignore the imports from distutils
to fix the mypy build.

### DIFF
--- a/pegen/build.py
+++ b/pegen/build.py
@@ -3,8 +3,8 @@ import shutil
 
 import distutils.log
 from distutils.core import Distribution, Extension
-from distutils.command.clean import clean
-from distutils.command.build_ext import build_ext
+from distutils.command.clean import clean # type: ignore
+from distutils.command.build_ext import build_ext # type: ignore
 
 MOD_DIR = pathlib.Path(__file__)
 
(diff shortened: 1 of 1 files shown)


gvanrossum:
Are you submitting a fix to typeshed too?

pablogsal:
> Are you submitting a fix to typeshed too?

Will probably do sometime this week, as the fix is probably bigger in scope because typeshed has little stubs for disutils.command. In the worst case will open an issue so I remember to do it in the future.

gvanrossum:
Do you know of stubgen? It's a script that comes with mypy, it can generate
stubs automatically that are pretty decent.

pablogsal:
> Do you know of stubgen? It's a script that comes with mypy, it can generate
stubs automatically that are pretty decent.

Oh, I did not knew about that, thanks for pointing that out! I will try it soon and see if what I can get from that.

PR #13, opened 2019-08-05
Add makefile targets as tests in Travis-CI

pablogsal:
Not sure if this is something we want for an extended period of time, but I though it may be useful to check that nothing breaks the makefile targets meanwhile we extend the test suite.


pablogsal:
This PR needs #12 to land first

PR #14, opened 2019-08-05
Add some improvements to the C generated parser

pablogsal:
* Add a function to allow the parser to parse python strings directly.
* Add free code to the run_parser function and new specialized functions.
* Make the Makefile default target depend on pegen.c and headers (so it regenerates when editing them).
* Speed up the c parser test by generating the extension only once.
* Move to use the Python allocator to make it easier to trace stuff in the future.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -13,13 +13,17 @@ clean:
 
 dump: pegen/parse.c
 	cat -n $(TESTFILE)
-	$(PYTHON) -c "from pegen import parse; import ast; t = parse.parse('$(TESTFILE)'); print(ast.dump(t))"
+	$(PYTHON) -c "from pegen import parse; import ast; t = parse.parse_file('$(TESTFILE)'); print(ast.dump(t))"
+
+# Note: These targets really depend on the generated shared object in pegen/parse.*.so but
+# this has different names in different systems so we are abusing the implicit dependency on
+# parse.c by the use of --compile-extension.
 
 test: pegen/parse.c
-	$(PYTHON) -c "from pegen import parse; import ast; t = parse.parse('$(TESTFILE)'); exec(compile(t, '', 'exec'))"
+	$(PYTHON) -c "from pegen import parse; t = parse.parse_file('$(TESTFILE)'); exec(compile(t, '', 'exec'))"
 
 time: pegen/parse.c
-	/usr/bin/time -l $(PYTHON) -c "from pegen import parse; parse.parse('$(TIMEFILE)')"
+	/usr/bin/time -l $(PYTHON) -c "from pegen import parse; parse.parse_file('$(TIMEFILE)')"
 
 time_stdlib:
 	/usr/bin/time -l $(PYTHON) -c "import ast; ast.parse(open('$(TIMEFILE)').read())"
--- a/Makefile
+++ b/Makefile
@@ -5,7 +5,7 @@ GRAMMAR = data/cprog.gram
 TESTFILE = data/cprog.txt
 TIMEFILE = data/xxl.txt
 
-pegen/parse.c: $(GRAMMAR) pegen/*.py
+pegen/parse.c: $(GRAMMAR) pegen/*.py pegen/pegen.c pegen/*.h
 	$(PYTHON) -m pegen -q -c $(GRAMMAR) -o pegen/parse.c --compile-extension
 
 clean:
--- a/pegen/c_generator.py
+++ b/pegen/c_generator.py
@@ -35,8 +35,6 @@ parse_string(PyObject *self, PyObject *args)
     return run_parser_from_string(the_string, (void *)start_rule, %(mode)s);
 }
 
-
-
 static PyMethodDef ParseMethods[] = {
     {"parse_file",  parse_file, METH_VARARGS, "Parse a file."},
     {"parse_string",  parse_string, METH_VARARGS, "Parse a string."},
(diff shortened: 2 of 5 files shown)


Makefile
```diff
@@ -5,7 +5,7 @@ GRAMMAR = data/cprog.gram
 TESTFILE = data/cprog.txt
 TIMEFILE = data/xxl.txt
 
-pegen/parse.c: $(GRAMMAR) pegen/*.py
+pegen/parse.c: $(GRAMMAR) pegen/*.py pegen/pegen.c pegen/*.h
 	$(PYTHON) -m pegen -q -c $(GRAMMAR) -o pegen/parse.c --compile-extension
```
gvanrossum: I just realized this is abusing make's dependency mechanism. The make rule on this line uses `pegen/parse.c` as the target, but the targets that depend on it (`dump`, `test`, `time`) all need the *side effect* of using `--compile-extension`, which produces a `.so` file.

On my Mac that `.so` file is called `pegen/parse.cpython-38-darwin.so`. If it's also called that on Linux, let's name that file instead of `pegen/parse.c` as the target here and as the dependency of the other targets.

If it's got a different name on Linux, let's keep it this way, but let's add a comment.

Makefile
```diff
@@ -5,7 +5,7 @@ GRAMMAR = data/cprog.gram
...
 
 clean:
```
gvanrossum: Note for line 19 below: the `test` target needn't import `ast` -- that was a copy/paste from the `dump` target.

pegen/c_generator.py
```diff
@@ -22,11 +22,22 @@
...
+
+    if (!PyArg_ParseTuple(args, "s", &the_string))
+        return NULL;
+    return run_parser_from_string(the_string, (void *)start_rule, %(mode)s);
 }
 
 static PyMethodDef ParseMethods[] = {
-    {"parse",  parse_file, METH_VARARGS, "Parse a file."},
+    {"parse_file",  parse_file, METH_VARARGS, "Parse a file."},
+    {"parse_string",  parse_string, METH_VARARGS, "Parse a string."},
```
gvanrossum: If you're changing the public name of the function, you should also change the lines where the function is called in the `Makefile`.

pegen/pegen.c
```diff
@@ -221,44 +221,81 @@ keyword_token(Parser *p, const char *val)
...
+    p->tokens = PyMem_Malloc(sizeof(Token));
     memset(p->tokens, '\0', sizeof(Token));
     p->mark = 0;
     p->fill = 0;
     p->size = 1;
 
     p->arena = PyArena_New();
-    if (!p->arena)
-        return NULL;
+    if (!p->arena){
```
gvanrossum: Space between ')' and '{'.

pegen/c_generator.py
pablogsal: Yep, I realized afterwards in the Travis-CI failure thanks to #13 (That payed off very quickly) :D

Makefile
pablogsal: Yeah, you guessed the reason for this "hack" and why I used the implicit dependency on `--compile-extension`: on Linux this is called `parse.cpython-38-x86_64-linux-gnu.so` (and this can change in other platforms). I will add a comment to explain that.

pablogsal:
> Next up, getting rid of (most) panic() calls?

Yeah I think that is the next logical step :)

pegen/pegen.c
```diff
@@ -221,44 +220,88 @@ keyword_token(Parser *p, const char *val)
...
+        fail = 1;
+        goto exit;
+    }
+
+exit:
+
+    PyMem_Free(p->tokens);
+
+    if (fail) {
+        PyArena_Free(p->arena);
```
pablogsal: There is a bit of redundancy here on the cleanup for the arena and p, but I find this more readable than other versions that I tried factoring that out. Tell me if you have a preference in mind or some other structure that I have not consider.

pegen/pegen.c
```diff
@@ -221,44 +220,88 @@ keyword_token(Parser *p, const char *val)
...
+        PyMem_Free(p);
         return NULL;
     }
 
-    // TODO: Free stuff
-    if (mode == 1)
-        return PyAST_mod2obj(res);
+    if (mode == 1){
```
gvanrossum: pep7

pegen/pegen.c
gvanrossum: It's always "arena free; mem free; return" -- why not have a shared `result` variable and a `goto`?

pablogsal: Indeed, I will take this as a sign that is too late for today :sweat_smile:

pablogsal:
Fixed PEP7 and the error handling. I have added a new `panic`call for a `PyMem_Malloc` fail path so I remember to address that when we remove the panics.

pegen/pegen.c
```diff
@@ -246,30 +247,24 @@ run_parser(struct tok_state* tok, void *(start_rule_func)(Parser *), int mode)
     void *res = (*start_rule_func)(p);
     if (res == NULL) {
         PyErr_Format(PyExc_SyntaxError, "error at mark %d, fill %d, size %d", p->mark, p->fill, p->size);
-        fail = 1;
         goto exit;
     }
 
+    if (mode == 1) {
+        result =  PyAST_mod2obj(res);
```
gvanrossum: FWIW I would love it if there was a mode that returned the compiled bytecode instead of the AST. That would allow a purer comparison with the current compiler (I believe that converting the internal AST to the public AST costs more than the rest of parsing or compilation). That's also something for another day...

pablogsal: At some point it may be beneficial to use GitHub issues to make sure we do not forget to do these and to make sure we don't duplicate work.

PR #19, opened 2019-08-12
Remove panic calls in the c parser generator

pablogsal:


pegen/c_generator.py
```diff
@@ -145,6 +146,31 @@ class CParserGenerator(ParserGenerator, GrammarVisitor):
...
+        self._varname_counter = itertools.count()
+
+    def unique_varname(self, name="tmpvar"):
+        return name + "_" + str(next(self._varname_counter))
+
+    def call_with_errorcheck(self, call_text, returnval="", goto_target=None):
+        error_var = self.unique_varname()
+        self.print(f"int {error_var} = {call_text}")
+        self.print(f"if ({error_var}) {{")
+        with self.indent():
```
pablogsal: Not super fan of this way of adding error handling, I want to think about if there is a better abstraction. Suggestions welcome :)

pegen/c_generator.py
```diff
@@ -145,6 +146,31 @@ class CParserGenerator(ParserGenerator, GrammarVisitor):
     def __init__(self, rules: Dict[str, grammar.Rule], file: Optional[IO[Text]]):
         super().__init__(rules, file)
         self.callmakervisitor = CCallMakerVisitor(self)
+        self._varname_counter = itertools.count()
+
+    def unique_varname(self, name="tmpvar"):
+        return name + "_" + str(next(self._varname_counter))
```
pablogsal: I think we may want to use this more often when generating temporary varnames.

pegen/c_generator.py
```diff
@@ -145,6 +146,31 @@ class CParserGenerator(ParserGenerator, GrammarVisitor):
...
+            if goto_target:
+                self.print(f"goto {goto_target};")
+            else:
+                self.print(f"return {returnval};")
+        self.print(f"}}")
+
+    def out_of_memory(self, expr, text="Parser out of memory", returnval="", goto_target=None):
+        self.print(f"if ({expr}) {{")
+        with self.indent():
+            self.print(f'PyErr_Format(PyExc_MemoryError, "{text}");')
```
gvanrossum: Maybe I'm prematurely optimizing, but I wonder if this warrants a tiny C-level helper that you can call without arguments that raises MemoryError with a fixed error string. Exactly where it ran out of memory is unlikely to be of interest (and if you really care you can use gdb/lldb).

pegen/c_generator.py
```diff
@@ -145,6 +146,31 @@ class CParserGenerator(ParserGenerator, GrammarVisitor):
...
+        self.print(f"int {error_var} = {call_text}")
+        self.print(f"if ({error_var}) {{")
+        with self.indent():
+            if goto_target:
+                self.print(f"goto {goto_target};")
+            else:
+                self.print(f"return {returnval};")
+        self.print(f"}}")
+
+    def out_of_memory(self, expr, text="Parser out of memory", returnval="", goto_target=None):
```
gvanrossum: I'd rather have separate helpers to generate the `goto` version vs. the `return` version. Then make the return value not have a default. Ditto above.

pegen/c_generator.py
```diff
@@ -145,6 +146,31 @@ class CParserGenerator(ParserGenerator, GrammarVisitor):
...
         super().__init__(rules, file)
         self.callmakervisitor = CCallMakerVisitor(self)
+        self._varname_counter = itertools.count()
+
+    def unique_varname(self, name="tmpvar"):
+        return name + "_" + str(next(self._varname_counter))
+
+    def call_with_errorcheck(self, call_text, returnval="", goto_target=None):
+        error_var = self.unique_varname()
+        self.print(f"int {error_var} = {call_text}")
```
gvanrossum: I find it jarring that the `;` has to be passed in -- maybe generate it here and remove it from the call site.

pegen/c_generator.py
gvanrossum: Agreed, though I am not very familiar with `itertools.count()` so I'd rather use a simple counter that's updated here manually.

pablogsal:
Humm...it seems that the CI is failing because one test in story 4 introduced in commit 02e534279320d00ca2043494e78de762923d2063 is failing (also on `master`):

 ```exception
    def test_toy():
        program = "x - (y + z)"
        file = StringIO(program)
        tokengen = generate_tokens(file.readline)
        tok = Tokenizer(tokengen)
        p = ToyParser(tok)
        tree = p.statement()
        print(tree)
        assert tree and tree.type == "statement"
        assert tree.children[0].type == "expr"
>       assert tree.children[0].children[0].type == "term"
E       AssertionError: assert 'expr' == 'term'
E         - expr
E         + term
```

File: pegen/c_generator.py (hunk 3 above)
Comment (pablogsal) [reply pegen/c_generator.py]: You mean a function that just calls `PyErr_Format(PyExc_MemoryError, "The parser ran out of memory");` or a  macro that also returns?

Comment (gvanrossum) [reply pegen/c_generator.py]: I meant a function. It's premature because I'm trying to reduce binary code size in a prototype. (I do expect that there will be a lot of generated code. But still I'm being premature.)

File: pegen/c_generator.py [hunk 6]
```diff
@@ -225,8 +236,8 @@ def visit_Rule(self, node):
                 self.print("int resmark = p->mark;")
                 self.print("while (1) {")
                 with self.indent():
-                    self.call_with_errorcheck(
-                            f"update_memo(p, mark, {node.name}_type, res);",
+                    self.call_with_errorcheck_return(
+                            f"update_memo(p, mark, {node.name}_type, res)",
                             returnval="res"
```
gvanrossum: I'd use a positional arg for returnval everywhere.

pegen/c_generator.py
```diff
@@ -146,30 +146,41 @@ class CParserGenerator(ParserGenerator, GrammarVisitor):
...
+    def call_with_errorcheck_goto(self, call_text, goto_target):
+        error_var = self.unique_varname()
+        self.print(f"int {error_var} = {call_text};")
+        self.print(f"if ({error_var}) {{")
+        with self.indent():
+            self.print(f"goto {goto_target};")
         self.print(f"}}")
 
-    def out_of_memory(self, expr, text="Parser out of memory", returnval="", goto_target=None):
+    def out_of_memory_return(self, expr, returnval, text="Parser out of memory"):
```
gvanrossum: Rename `text` to `message`?

PR #22, opened 2019-09-10
Allow passing a grammar file path to the directory test script to regenerate the parser 

emilyemorehouse:
- Add a helper function for generating the parser, with error handling if an error occurs
- Add a grammar file arg
- Update `sys.exit` calls to supply error code
- Refactor parser generation into a helper function. Update pegen/__main__.py and test_parse_directory.py to use the new helper
- Remove `endpos`

### DIFF
--- a/test_parse_directory.py
+++ b/test_parse_directory.py
@@ -1,7 +1,14 @@
 import argparse
 import os
-import pegen
 import sys
+import tokenize
+import traceback
+
+from pegen.build import compile_c_extension
+from pegen.c_generator import CParserGenerator
+from pegen.tokenizer import Tokenizer
+from pegen.tokenizer import grammar_tokenizer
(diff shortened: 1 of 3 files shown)


test_parse_directory.py
```diff
@@ -37,19 +45,56 @@ def report_status(succeeded, file, verbose, error=None):
         print(f"  {str(error.__class__.__name__)}: {error}")
 
 
+def generate_parser(grammar_file):
```
gvanrossum: Can you factor this out into a utility that's also called from `__main__.py`? It doesn't need to set or return `endpos` -- we can just drop that from the output (it was more useful before I had a line count, to estimate the compilation speed).

The helper function could return the exception object representing the SyntaxError, or None for success.

emilyemorehouse:
I'm working on fixing the errors 😬

PR #24, opened 2019-09-11
Refactor c-parser tests into its own file and add some simple test cases for PEG features

pablogsal:


test/test_c_parser.py
```diff
@@ -0,0 +1,79 @@
+import ast
+
```
emilyemorehouse:
```suggestion
```

test/test_c_parser.py
```diff
@@ -0,0 +1,79 @@
...
+import pytest
+
+from pegen.grammar import GrammarParser
+
```
emilyemorehouse:
```suggestion
```

test/test_c_parser.py
```diff
@@ -0,0 +1,79 @@
...
+    expr: NAME | NUMBER
+    """
+
+    rules = parse_string(grammar, GrammarParser).rules
+    extension = generate_parser_c_extension(rules, tmp_path)
+
+    extension.parse_string("foo bar")
+
+    with pytest.raises(SyntaxError):
+        extension.parse_string("foo 34")
```
emilyemorehouse: Since I expect we'll be adding quite a few more tests that use the same form as `test_lookahead` and  `test_negative_lookahead`, it may be helpful to abstract this into another helper that we can reuse. Possibly something we can pass the grammar to along with a list of success and failure cases?

test/test_c_parser.py
```diff
@@ -0,0 +1,79 @@
+import ast
+import pytest
+
+from pegen.grammar import GrammarParser
+from test.util import parse_string, import_file, generate_parser_c_extension
+
+
+def check_input_strings_for_grammar(grammar, tmp_path, valid_cases=[], invalid_cases=[]):
```
gvanrossum: Mutable defaults are likely to get lint errors. :-)

test/test_pegen.py
```diff
@@ -11,110 +8,8 @@
...
-        "(1+1) - (1+1)",
-        "(1+1) * (1+1)",
-        "(1+1) / (1+1)",
-    ]
-
-    for expr in expressions:
-        the_ast = extension.parse_string(expr)
-        expected_ast = ast.parse(expr)
-        assert ast.dump(the_ast) == ast.dump(expected_ast)
+from test.util import generate_parser, parse_string, make_parser
```
gvanrossum: Hm... I'm not excited to see the test directory being used as a package namespace -- there's a stdlib package `test` that sometimes is imported for stuff, and this would prevent that. Maybe move the test utilities into `pegen` (named `testutil`)?

test/util.py
```diff
@@ -0,0 +1,70 @@
...
+    source = path / "parse.c"
+    with open(source, "w") as file:
+        genr = CParserGenerator(rules, file)
+        genr.generate("parse.c")
+    extension_path = compile_c_extension(str(source),
+                                         build_dir=str(path / "build"))
+    extension = import_file("parse", extension_path)
+    return extension
+
+
```
gvanrossum: Trailing blank lines.

PR #25, opened 2019-09-11
Implement cut operator on the C generator

pablogsal:


pegen/c_generator.py
```diff
@@ -374,18 +376,21 @@ def add_var(self, node, names: List[str]) -> Tuple[str, Optional[str]]:
...
-                    if rule.is_loop():
-                        type = 'asdl_seq *'
-                    else:
-                        type = rule.type
-                elif name.startswith('_loop'):
+        if not name:
+            return name, type
+        if name.startswith('cut'):
+            return name, 'int'
+        if name.endswith('_var'):
```
pablogsal: This looks like a huge change but is just a dedent (as we are handling 'cut' and no name before).

PR #26, opened 2019-09-11
Make test_parse_directory script exit with error if there is any error

pablogsal:


pegen/build.py
```diff
@@ -7,6 +7,7 @@
 from distutils.command.clean import clean  # type: ignore
 from distutils.command.build_ext import build_ext  # type: ignore
 
+from pegen.python_generator import PythonParserGenerator
```
pablogsal: This is a missing import (the main script fails to generate the python parser currently) that is unrelated to this PR but was introduced in the same change.

PR #28, opened 2019-09-11
Refactor build methods to separate parser and generator. Use helper to reduce code in grammar_visualizer and rename methods accordingly.

emilyemorehouse:


pegen/build.py
```diff
@@ -100,4 +91,41 @@ def build_parser(
...
+    gen = build_generator(
+        tokenizer,
+        rules,
+        grammar_file,
+        output_file,
+        compile_extension,
+        verbose_c_extension,
+    )
+
     return rules, parser, tokenizer, gen
```
gvanrossum: Hm... It's kind of confusing that here we return `rules, parser, ...` whereas `build_parser()` returns `parser, rules, ...`. Can you make it so that here we just return an extra element but the first three are in the same order? (I honestly don't care which order, as long as it's consistent. :-)

emilyemorehouse: Ah yes, good catch. Done.

PR #34, opened 2019-09-13
Add helper script for finding the level at which the parsers will fail for nested expressions

emilyemorehouse:


scripts/find_max_stacksize.py
```diff
@@ -0,0 +1,83 @@
...
+        print(f"({language}) Nesting depth of {nesting_depth} is successful")
+
+        return True
+    except Exception as err:
+        print(f"{FAIL}({language}) Failed with nesting depth of {nesting_depth}{ENDC}")
+        print(f"{FAIL}\t{err}{ENDC}")
+        return False
+
+
+if __name__ == "__main__":
```
gvanrossum: I'd create a main() function and call that under this condition. Otherwise LG!

PR #37, opened 2019-09-16
Bootstrap the grammar parser (generate the grammar parser from a metagrammar file)

pablogsal:
This PR implements the set of *minimal changes* needed to automatically regenerate the grammar parser. 

I left for other PRs/ future work:
   * Implementing the possibility of adding headers (`@` notation).
   * The work needed to get rid of the `CURLY_STUFF`.

The reason for implementing just the *minimal changes* is that the bootstrap process is very delicate and iterating over this while maintaining the parse tree for the generators unaltered can be quite tricky.


pegen/testutil.py
```diff
@@ -70,7 +70,7 @@ def generate_parser_c_extension(rules, path):
 
 
 def print_memstats() -> bool:
-    MiB: Final = 2 ** 20
```
pablogsal: This is needed because master is currently failing due to this type being unavailable.

gvanrossum: Oh actually the right fix is to add `from __future__ import annotations` to the top of the file.

pegen/grammar.py
```diff
@@ -37,7 +37,7 @@ def generic_visit(self, node, *args, **kwargs):
 class Rules:
```
gvanrossum: (We should really rename this class to Grammar -- but not in this PR.)

pegen/grammar.py
```diff
@@ -386,192 +386,22 @@ def initial_names(self) -> AbstractSet[str]:
...
+        if not isinstance(other, Cut):
+            return NotImplemented
+        return True
+
+
 Plain = Union[Leaf, Group]
-Item = Union[Plain, Opt, Repeat, Lookahead, Rhs]
-
-
-class GrammarParser(Parser):
```
gvanrossum: In story6 I kept the old grammar parser around just so that I could still parse the meta-grammar even if I had screwed up something that broke the generated meta-parser. Check out the `-b` flag there.

pegen/grammar.py
```diff
@@ -386,192 +386,22 @@ def initial_names(self) -> AbstractSet[str]:
...
-            return Group(alts)
-        self.reset(mark)
-        if name := self.name():
-            assert name
-            return NameLeaf(name.string)
-        if string := self.string():
-            assert string
-            return StringLeaf(string.string)
-        return None
+Item = Union[Plain, Opt, Repeat, Lookahead, Rhs]
```
gvanrossum: Can you configure your editor to always ensure a file ends in \n (unless empty)?

pegen/metagrammar.gram
```diff
@@ -0,0 +1,55 @@
+start: rules ENDMARKER {Rules(rules)}
+
+rules:
+    | rule rules { [rule] + rules }
+    | rule { [rule] }
+
+rule:
+    | rulename ":" alts NEWLINE INDENT more_alts DEDENT { Rule(*rulename, alts + more_alts) }
```
gvanrossum: I betcha this line doesn't work -- it's trying to add two `Rhs` instances.

I recommend just dropping it for now.

pegen/metagrammar.gram
```diff
@@ -0,0 +1,55 @@
...
+    | rulename ":" NEWLINE INDENT more_alts DEDENT { Rule(*rulename, more_alts) }
+    | rulename ":" alts NEWLINE { Rule(*rulename, alts) }
+
+rulename:
+        | NAME '[' type=NAME '*' ']' {(name.string, type.string+"*")}
```
gvanrossum: Just use 4-space indents for everything. :-)

pegen/metagrammar.gram
```diff
@@ -0,0 +1,55 @@
...
+alts:
+    | alt "|" alts { Rhs([alt] + alts.alts)}
+    | alt { Rhs([alt]) }
+
+more_alts:
+    | "|" alts NEWLINE more_alts { Rhs(alts.alts + more_alts.alts) }
+    | "|" alts NEWLINE { Rhs(alts.alts) }
+
+alt:
+    | items ['$'] action { Alt(items + ([NamedItem(None, NameLeaf('ENDMARKER'))] if opt else []), action=action.string) }
```
gvanrossum: I would give an explicit name to the `['$']` piece.

Also the action is hard to follow -- maybe just write out two alternatives, `item '$' action` and `item action`? (And two more for the variation without `action`.)

pegen/testutil.py
gvanrossum: Alas, that's not true, mypy will still complain. The true fix is to add
```
from typing_extensions import Final
```

pegen/grammar.py
pablogsal: I can do this, but one thing to notice is that that with the new rules (especially the possibility to write every option in a new line) the old grammar parser cannot parse the metagrammar. This is why I have eliminated it completely.

pegen/metagrammar.gram
pablogsal: I can do the same as I did in `more_alts`:

```
Rhs(alts.alts + more_alts.alts)
```

pegen/grammar.py
gvanrossum: OK, that's fine. I think for story6 I updated the hand-written grammar parser, but since we use Git we don't really need a backup anyway... :-)

pegen/metagrammar.gram
gvanrossum: Sure.

PR #39, opened 2019-09-16
Rename Grammar -> Rules and use Grammar node instead of a dictionary

pablogsal:


pegen/testutil.py
```diff
@@ -46,7 +46,7 @@ def parse_string(source, parser_class, *, dedent=True, verbose=False):
 
 def make_parser(source):
     # Combine parse_string() and generate_parser().
-    rules = parse_string(source, GrammarParser).rules
+    rules = parse_string(source, GrammarParser)
```
gvanrossum: I think this returns a `Grammar` so the variable should be named `grammar`.

test/test_pegen.py
```diff
@@ -531,7 +535,7 @@ def test_parse_or_grammar(self):
 
         visitor.visit(rules)
 
-        # Rules/Rule/Rhs/Alt/NamedItem/NameLeaf   -> 6
+        # Grammar/Rule/Rhs/Alt/NamedItem/NameLeaf   -> 6
```
gvanrossum: Fix ASCII art alignment?

pegen/build.py
```diff
@@ -79,9 +79,9 @@ def build_generator(
     with open(output_file, "w") as file:
         gen: ParserGenerator
         if output_file.endswith(".c"):
-            gen = CParserGenerator(rules.rules, file)
+            gen = CParserGenerator(rules, file)
```
gvanrossum: Looks like this argument should be called `grammar`. (Also in the `def` line above and for `PythonParserGenerator` below).

pablogsal:
Thanks for the fixes!

PR #40, opened 2019-09-16
 Remove CURLY_STUFF hack 

pablogsal:


pablogsal:
Note: This PR is rebased on top of #39 (it has some commits of that PR).

pegen/grammar_parser.py
```diff
@@ -28,7 +28,7 @@ class GeneratedParser(Parser):
 
     @memoize
     def start(self):
-        # start: rules $ { Grammar ( rules ) }
+        # start: rules $ {'Grammar ( rules )'}
```
gvanrossum: I don't like this change -- there must be something in the generator that causes this change.

pegen/metagrammar.gram
```diff
@@ -54,4 +54,17 @@ atom:
     | NAME {NameLeaf(name.string) }
     | STRING {StringLeaf(string.string)}
 
-action: action=CURLY_STUFF {action}
+# Mini-grammar for the actions
```
gvanrossum: Reminder to update TODO.md.

pegen/metagrammar.gram
```diff
@@ -54,4 +54,17 @@ atom:
...
+
+action: "{" stuffs "}" { stuffs }
+
+stuffs:
```
gvanrossum: Can you think of better names than `stuffs` and `stuff`? I meant them to be like `foo` and `bar` but never got around to inventing better names. Now's the time, I guess. :-)

pablogsal: I updated the PR to use the uninspired `action_items`  and `action_item`. I was thinking about something with `atom`/`molecule` but I was unsure if it fits the convention with those. 

Please, advice if you find better names :)

PR #42, opened 2019-09-19
Add types to generated parser

gvanrossum:
But turn strict optional off for that file because mypy doesn't get that 'if a := b:' implies 'if a:'.

### DIFF
--- a/mypy.ini
+++ b/mypy.ini
@@ -23,5 +23,4 @@ show_traceback = True
 show_error_codes = True
 
 [mypy-pegen.grammar_parser]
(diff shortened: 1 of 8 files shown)


mypy.ini
```diff
@@ -6,7 +6,7 @@ no_implicit_optional = True
 strict_optional = True
 
 #check_untyped_defs = True
-#disallow_untyped_calls = True
+disallow_untyped_calls = True
```
gvanrossum: Had meant to push this one separately, but oh well.

PR #53, opened 2019-10-16
Validate implementation of 'cut' expressions

### LINKED ISSUE #49: Validate implementation of the 'cut' operator
The 'cut' operator serves two purposes in PEG (and other types of grammars):

1. It commits the parser to the current alternative, so further alternatives will not be tried.

The above is important because a parser must deal with valid programs, and _also_ with not-valid ones. The `cut` operator is the best way to make the parser commit to failure fast (an intentionally ambiguous pseudo-Python source could make the parser backtrack a lot).

2. It allows to prune the memoization cache, saving memory.

apalala:
There needs to be a _'cut' context_ for each choice (`|`), optional (`[e]`), and closure (`e*`, `e+`), so a cut (`~`) expression will commit the most recent context at runtime.

fixes #49

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,5 +1,5 @@
 TOP = /usr/local
-PYTHON ?= $(TOP)/bin/python3.8
+PYTHON ?= `which python3.8`
 
 GRAMMAR = data/cprog.gram
 TESTFILE = data/cprog.txt
--- a/data/cprog.gram
+++ b/data/cprog.gram
@@ -1,14 +1,14 @@
 start[mod_ty]: a=stmt* ENDMARKER { Module(a, NULL, p->arena) }
 stmt[stmt_ty]: compound_stmt | simple_stmt
-compound_stmt[stmt_ty]: pass_stmt | if_stmt
+compound_stmt[stmt_ty]: if_stmt | pass_stmt
 
-pass_stmt[stmt_ty]: a='pass' NEWLINE { _Py_Pass(EXTRA(a, a)) }
+pass_stmt[stmt_ty]: a='pass' ~ NEWLINE { _Py_Pass(EXTRA(a, a)) }
 
 if_stmt[stmt_ty]:
     | 'if' ~ c=expr ':' t=suite e=[else_clause] { _Py_If(c, t, e, EXTRA(c, c)) }
 else_clause[asdl_seq*]:
-    | 'elif' c=expr ':' t=suite e=[else_clause] { singleton_seq(p, _Py_If(c, t, e, EXTRA(c, c))) }
-    | 'else' ':' s=suite { s }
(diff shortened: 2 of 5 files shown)


apalala:
A problem with the current grammar model, and one that will surface with grouping (`(` `)`) is that the `Rhs` type and the generator doesn't provide different treatment for expression sequences, choices, optionals, and closures. It would be best to have one `visit_X` method for each, even if there is some repetition (copy&paste) in `c_generator`. Saving some lines with `is_loop` doesn't seem that good when implementing 'cut' that way would require an `is_choice` or the likes.

In short, What is `Rhs`? The major `Node` subtypes should be  `[Seq, Choice, Opt, Repeat0, Repeat1, Group]`, with the grammar parser transforming single-alt `Choice` to `Seq`, and single-item `Seq` to `items[0]`. Note that `Group.node` may be `Choice`, `Seq`, etc. This is not a big change, and the sooner it gets done the better.

apalala:
@gvanrossum, I'd remove the automatic generation of names for RHS terms now that you implemented `name=exp`. The automated name generation seems to be the only reason to have a `CCallMakerVisitor` and two passes over the grammar tree, and every `exp` having a name complicates the code generator. 

On the same vain, I don't see the need for two kinds of `NamedXXX` (`NamedItem`, `NameLeaf`) if `Named` is reserved for `name=exp`. 

Most parts of the RHS of a grammar rule are discardable, so generating names for them at the expense of complicating the code generator is not worth it.

(p.s. this doesn't seem like the right place for these comments, but...)

gvanrossum:
The failures seem to point to something amiss in the changes.

I really wish we could start committing pegen/parse.c -- then it would be clearer how the generated code changes (I had it that way, but during the core dev sprints we decided to stop checking in generated code. I now regret this decision.)

I'm still studying these changes.

gvanrossum:
PS. I really would like to make corresponding changes in the Python generator too (or even first).

apalala:
The important changes (the refactoring) will affect the grammar model (no more `Rhs`), so `python_generator` must be updated.

I created this pull request just to discuss the ideas. The approach cannot work without the refactoring, and will probably not work if the generated code is to remain simple.

I'll open other issues for the refactoring, which should take about 1/2 a day to implement for someone familiar with the code, and probably 1.5 days for me. Do you have to tackle this week, @gvanrossum ?

gvanrossum:
OK, let's close this PR then and instead work on the refactoring. We can restart this after we have agreement on the semantics of *cut*.

gvanrossum:
(You should be able to close your own PR. Let me know if you can't.)

lysnikolaou:
I feel that this can be closed, since 1,5 months have passed since the last bit of activity. @gvanrossum I don't feel comfortable enough to close this myself, so would you do it if it's okay with you?

gvanrossum:
Yeah, this isn’t going anywhere. Sorry @apalala.

PR #60, opened 2019-10-27
Refactor Simpy Grammar and add helper functions to generate AST for pass

lysnikolaou:
These changes correctly parse and compile `pass` into an AST.

Due to the *partial* refactoring of `simpy.gram`, the generated parser does not parse the `cpython/Lib` folder correctly anymore, rather it crashes with a Segmentation Fault. I don't really know how we want to proceed here.

Another slight problem: I know that the `test/test_data` thing is a somewhat dumb way of doing things for such small python files, but this is only a start and as we refactor more of `simpy.gram`, this will be a nice place to have python source files to be tested.

And finally, my `C` changes might not adhere to the CPython coding style, but it's only my first time and I'm of course happy to change whatever is not correct.


lysnikolaou:
Also, if we decide to go with the `test/test_data` directory for test python files, it might be a good idea to get all the files with something like `pathlib`'s `Path.iterdir()` instead of listing them manually.

gvanrossum:
I am assuming the problem is due to `CONSTRUCTOR()` (which returns `(void*)1`) but I'm not yet clear on exactly how, or how to fix it. You may end up having to compile CPython from source with debugging on and run a test under a debugger (gdb on Linux, or lldb on Mac -- which are you using?)

lysnikolaou:
That might indeed be the problem, although I can't figure out why would refactoring these rules would make a difference. I already have compiled CPython with debugging on on my Mac machine, where I use lldb. I'm going to test further under lldb tomorrow.

pegen/pegen.c
```diff
@@ -432,3 +432,53 @@ singleton_seq(Parser *p, void *a)
...
+    }
+
+    int flattened_seq_idx = 0;
+    for (int i = 0, l = asdl_seq_LEN(seqs); i < l; i++) {
+        asdl_seq *inner_seq = asdl_seq_GET(seqs, i);
+        for (int j = 0, li = asdl_seq_LEN(inner_seq); j < li; j++) {
+            asdl_seq_SET(flattened_seq, flattened_seq_idx++, asdl_seq_GET(inner_seq, j));
+        }
+    }
+
```
gvanrossum: Maybe add `assert(flattened_seq_idx == flattened_seq_size);` here? (Not that I doubt it, this all looks good. :-)

pegen/pegen.h
```diff
@@ -58,3 +58,5 @@ void *CONSTRUCTOR(Parser *p, ...);
 PyObject *run_parser_from_file(const char *filename, void *(start_rule_func)(Parser *), int mode);
 PyObject *run_parser_from_string(const char *str, void *(start_rule_func)(Parser *), int mode);
 asdl_seq *singleton_seq(Parser *, void *);
+asdl_seq *seq_insert_in_front(Parser *, void *, asdl_seq *);
+asdl_seq *seq_flatten(Parser *, asdl_seq *);
```
gvanrossum: Please make sure that every non-empty file ends with a newline character.

test/test_ast_generation.py
```diff
@@ -0,0 +1,31 @@
...
+def read_python_source(path):
+    with open(path, "r") as file:
+        source = file.read()
+    return source
+
+
+def test_ast_generation_on_source_files(tmp_path):
+    extension = create_tmp_extension(tmp_path)
+    for filename in PYTHON_SOURCE_FILENAMES:
+        source = read_python_source(f"test/test_data/{filename}")
```
gvanrossum: Please use `os.path.join()` to concatenate paths.

test/test_ast_generation.py
```diff
@@ -0,0 +1,31 @@
+import ast
+
+from pegen.grammar_parser import GeneratedParser as GrammarParser
+from pegen.testutil import parse_string, generate_parser_c_extension
+
+PYTHON_SOURCE_FILENAMES = ("pass.py", "multiple_pass.py")
```
gvanrossum: I'd rather use `[]` than `()` here.

test/test_data/multiple_pass.py
```diff
@@ -0,0 +1 @@
+pass; pass
```
gvanrossum: Maybe add another line with a single `pass`, to exercise the flattening function?

lysnikolaou:
I think I have an idea, why the SEGFAULT happens, although I am not sure that it is correct, because I didn't figure out how to test under a debugger. Regular testing shows that this is indeed a cause, but I'm not sure if it's the only one.

First of all, compound statement should return a `singleton_seq` now that the return type is `asdl_seq *`. The main reason though is that in `seq_flatten` and in `get_flattened_seq_size`, I tried to call `asdl_seq_LEN` on an `asdl_seq *` that could be `(void *) 1`, in case `CONSTRUCTOR` was called before.

Does that make sense to you? I know that the fix is a major hack, but I fear that we will have to come up with some of those, until all of `simpy.gram` has been refactored.

pegen/pegen.c
```diff
@@ -458,6 +458,7 @@ get_flattened_seq_size(asdl_seq *seqs)
     int size = 0;
     for (int i = 0, l = asdl_seq_LEN(seqs); i < l; i++) {
         asdl_seq *inner_seq = asdl_seq_GET(seqs, i);
+        if (asdl_seq_GET(inner_seq, 0) == (void *) 1) continue;
```
gvanrossum: I'd add a comment explaining this exclusion (good catch though!)

gvanrossum:
> `compound_statement` should return a `singleton_seq` now that the return type is `asdl_seq *`.

Yes, that's the fix! Good catch. See code review.

lysnikolaou:
Done!

PR #67, opened 2019-11-06
Refactor import rules to correctly generate AST for import_name

lysnikolaou:
This parses and generates the AST for `import_name` statements (correctly) without any segfaults. 

The only problem, which I have completely ignored until now, because I don't really know what the solution for it would be, is that a wrong AST gets generated when parsing `import a.b`, because only `b` gets identified. That has to do with the fact that the first argument to `_Py_alias` is the `attr` only, without any mention to the rest of the value. Some help would be needed here!

### DIFF
--- a/data/simpy.gram
+++ b/data/simpy.gram
@@ -50,8 +50,8 @@ import_from: ('from' (('.' | '...')* !'import' dotted_name | ('.' | '...')+)
               'import' ('*' | '(' import_as_names ')' | import_as_names))
 import_as_name: NAME ['as' NAME]
 dotted_as_name[alias_ty]:
-    | a=dotted_name 'as' b=NAME { _Py_alias(get_identifier_from_expr_ty(a), get_identifier_from_expr_ty(b), p->arena) }
-    | a=dotted_name { _Py_alias(get_identifier_from_expr_ty(a), NULL, p->arena) }
+    | a=dotted_name 'as' b=NAME { _Py_alias(((expr_ty) a)->v.Name.id, ((expr_ty) b)->v.Name.id, p->arena) }
+    | a=dotted_name { _Py_alias(((expr_ty) a)->v.Name.id, NULL, p->arena) }
 import_as_names: import_as_name (',' import_as_name)* [',']
 dotted_as_names[asdl_seq*]: a=dotted_as_name b=further_dotted_as_names* { seq_insert_in_front(p, a, b) }
 further_dotted_as_names[alias_ty]: ',' a=dotted_as_name { a }
--- a/data/simpy.gram
+++ b/data/simpy.gram
@@ -44,15 +44,19 @@ break_stmt: a='break' { _Py_Break(EXTRA(a, a)) }
 continue_stmt: a='continue' { _Py_Continue(EXTRA(a, a)) }
 
 import_stmt: import_name | import_from
-import_name: 'import' dotted_as_names
+import_name[stmt_ty]: a='import' b=dotted_as_names { _Py_Import(b, EXTRA(a, b)) }
 # note below: the ('.' | '...') is necessary because '...' is tokenized as ELLIPSIS
 import_from: ('from' (('.' | '...')* !'import' dotted_name | ('.' | '...')+)
               'import' ('*' | '(' import_as_names ')' | import_as_names))
 import_as_name: NAME ['as' NAME]
-dotted_as_name: dotted_name ['as' NAME]
+dotted_as_name[alias_ty]:
+    | a=dotted_name 'as' b=NAME { _Py_alias(get_identifier_from_expr_ty(a), get_identifier_from_expr_ty(b), p->arena) }
+    | a=dotted_name { _Py_alias(get_identifier_from_expr_ty(a), NULL, p->arena) }
 import_as_names: import_as_name (',' import_as_name)* [',']
-dotted_as_names: dotted_as_name (',' dotted_as_name)*
-dotted_name: NAME ('.' NAME)*
+dotted_as_names[asdl_seq*]: a=dotted_as_name b=further_dotted_as_names* { seq_insert_in_front(p, a, b) }
+further_dotted_as_names[alias_ty]: ',' a=dotted_as_name { a }
+dotted_name[expr_ty]: a=NAME b=further_dotted_name* { seq_to_dotted_name(p, seq_insert_in_front(p, a, b)) }
+further_dotted_name[expr_ty]: '.' a=NAME { a }
 
 if_stmt:
     | 'if' a=full_expression ':' b=block c=else_block { _Py_If(a, b, c, EXTRA(a, c)) }
--- a/data/simpy.gram
+++ b/data/simpy.gram
@@ -54,7 +54,7 @@ dotted_as_name[alias_ty]:
     | a=dotted_name 'as' b=NAME { _Py_alias(((expr_ty) a)->v.Name.id, ((expr_ty) b)->v.Name.id, p->arena) }
     | a=dotted_name { _Py_alias(((expr_ty) a)->v.Name.id, NULL, p->arena) }
 dotted_name[expr_ty]:
-    | a=NAME '.' b=dotted_name { join_names_with_dot(p, a, b) }
+    | a=dotted_name '.' b=NAME { join_names_with_dot(p, a, b) }
     | NAME
 
 if_stmt:
--- a/data/simpy.gram
+++ b/data/simpy.gram
@@ -54,7 +54,7 @@ dotted_as_name[alias_ty]:
     | a=dotted_name 'as' b=NAME { _Py_alias(((expr_ty) a)->v.Name.id, ((expr_ty) b)->v.Name.id, p->arena) }
     | a=dotted_name { _Py_alias(((expr_ty) a)->v.Name.id, NULL, p->arena) }
 dotted_name[expr_ty]:
-    | a=NAME b=('.' c=dotted_name { c }) { join_names_with_dot(p, a, b) }
+    | a=NAME '.' b=dotted_name { join_names_with_dot(p, a, b) }
     | NAME
 
 if_stmt:
--- a/data/simpy.gram
+++ b/data/simpy.gram
@@ -3,9 +3,8 @@
 start[mod_ty]: a=[statements] ENDMARKER { Module(a, NULL, p->arena) }
 statements[asdl_seq*]: a=statement+ { seq_flatten(p, a) }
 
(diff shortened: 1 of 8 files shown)


gvanrossum:
I'll give you a hint: in Python/ast.c, look at the flow in alias_for_import_name(), and see what it does for dotted_name.

lysnikolaou:
That really was a very helpful hint! Thanks!

lysnikolaou:
I dislike having to use something like `expr->v.Name.id` in so many places, do you think it would be a good idea to add a macro for that?

gvanrossum:
Hmm... Maybe. I would wait until you’ve written some more code.

I have some more review comments, will add those later.
-- 
--Guido (mobile)

data/simpy.gram
```diff
@@ -44,15 +44,19 @@ break_stmt: a='break' { _Py_Break(EXTRA(a, a)) }
...
+dotted_as_name[alias_ty]:
+    | a=dotted_name 'as' b=NAME { _Py_alias(((expr_ty) a)->v.Name.id, ((expr_ty) b)->v.Name.id, p->arena) }
+    | a=dotted_name { _Py_alias(((expr_ty) a)->v.Name.id, NULL, p->arena) }
 import_as_names: import_as_name (',' import_as_name)* [',']
-dotted_as_names: dotted_as_name (',' dotted_as_name)*
-dotted_name: NAME ('.' NAME)*
+dotted_as_names[asdl_seq*]: a=dotted_as_name b=further_dotted_as_names* { seq_insert_in_front(p, a, b) }
+further_dotted_as_names[alias_ty]: ',' a=dotted_as_name { a }
+dotted_name[expr_ty]: a=NAME b=further_dotted_name* { seq_to_dotted_name(p, seq_insert_in_front(p, a, b)) }
+further_dotted_name[expr_ty]: '.' a=NAME { a }
```
gvanrossum: Can you reorder the rules so they appear top-down?

Also `import_as_name[s]` should probably get different names, since they are components of `import_from`, not further variants of import statements. (If you get my drift. If you don't, let me know and I'll explain further.)

Finally, it should be possible to have actions inside parentheses, so you could get rid of `further_*`.

And an idea for later: I wonder if the pattern `X (sep X)*` is going to be so common that we may want to special-case this in the grammar, so we can return this as a sequence of Xs without needing to call `seq_insert_in_front()` (which makes an extra copy).

pegen/pegen.c
```diff
@@ -503,3 +503,64 @@ seq_flatten(Parser *p, asdl_seq *seqs)
...
+    // Get length of dotted string
+    ssize_t len = 0;
+    for (int i = 0, l = asdl_seq_LEN(seq); i < l; i++) {
+        expr_ty current_expr = asdl_seq_GET(seq, i);
+        ssize_t current_len = PyUnicode_GET_LENGTH(current_expr->v.Name.id);
+        len += current_len + 1;
+    }
+    --len; // Last name does not have a dot
+
+    PyObject *str = PyBytes_FromStringAndSize(NULL, len);
```
gvanrossum: I expect this won't be long enough if there are any non-ASCII characters in `seq`, which will cause the UTF8 encoding to take up more bytes than there are characters in the original.

pegen/pegen.c
```diff
@@ -503,3 +503,64 @@ seq_flatten(Parser *p, asdl_seq *seqs)
...
+
+    for (int i = 0, l = asdl_seq_LEN(seq); i < l; i++) {
+        expr_ty current_expr = asdl_seq_GET(seq, i);
+        const char *current_name = PyUnicode_AsUTF8(current_expr->v.Name.id);
+        if (!current_name) {
+            return NULL;
+        }
+        strcpy(s, current_name);
+        s += strlen(current_name);
+        *s++ = '.';
```
gvanrossum: I'm not happy about how this "overshoots" the target and then backs up. I'd rather only add a dot *before* all items except the first.

pegen/pegen.c
```diff
@@ -503,3 +503,64 @@ seq_flatten(Parser *p, asdl_seq *seqs)
...
+    *s = '\0';
+
+    PyObject *uni = PyUnicode_DecodeUTF8(PyBytes_AS_STRING(str),
+                                         PyBytes_GET_SIZE(str),
+                                         NULL);
+    Py_DECREF(str);
+    if (!uni) {
+        return NULL;
+    }
+    str = uni;
```
gvanrossum: This seems odd. Why not just keep using `uni`?

pegen/pegen.c
```diff
@@ -503,3 +503,64 @@ seq_flatten(Parser *p, asdl_seq *seqs)
...
+    if (PyArena_AddPyObject(p->arena, str) < 0) {
+        Py_DECREF(str);
+        return NULL;
+    }
+
+    expr_ty first_elem = asdl_seq_GET(seq, 0);
+    expr_ty last_elem = asdl_seq_GET(seq, asdl_seq_LEN(seq) - 1);
+    return Name(str, Load, first_elem->lineno, first_elem->col_offset,
+                last_elem->end_lineno, last_elem->end_col_offset, p->arena);
+}
```
gvanrossum: All in all this function looks very complex and error-prone. I wonder if we can't do better? One possibility would be to have just a function that takes two NAME tokens and joins them with a dot, and rewrite the grammar to be recursive instead of iterative. If there are a lot of dots that's more expensive, but I doubt that this would cost a measurable amount.

pegen/pegen.c
lysnikolaou: That's right. I wrongfully didn't think that through, thinking that it is not needed, since we are talking about modules and packages.

pegen/pegen.c
lysnikolaou: I agree that this version seems error-prone and that a recursive rule might work a bit better here. I was thinking about that, while I was writing it.

I am not sure if this is a valid concern, but in case we publish the grammar on the Docs like with the current one and we want to make it easier for people to understand it, making a recursive grammar for this might not be the best idea. In that case, I think that a better version of this function might just be good enough.

data/simpy.gram
lysnikolaou: I'll do the reordering, change the name for `import_as_name[s]` and try to remove all the unnecessary `further_*` rules.

On your last idea, quickly going through the grammar I found at least 16 occurrences of this pattern. Do you this it is enough to justify this special case in the grammar? I think, I'm more on the yes side.

gvanrossum: Great! Please open a separate issue to discuss the design of such a feature.

pegen/pegen.c
gvanrossum: I honestly think that right now we ought to concentrate on getting a valid AST for all Python programs. We'll rethink the display value of the grammar later. If you want to you create an issue to remind us to consider that separately.

gvanrossum:
PS. It's better not to force-push once code review is under way. Reasons:
1. GitHub doesn't send notifications
2. It's hard for the reviewer to tell what changed

lysnikolaou:
Oh okay, sorry about that! I only force-push, when I want to add something minor (most often something I have forgotten about) without adding a new commit and I need to rebase. Is it okay to just commit a little thing I may have missed in my previous commit?

pegen/pegen.c
lysnikolaou: Do you think I should open a new PR for this new recursive grammar or just commit to this branch?

gvanrossum:
It's not worth doing it -- all your commits will be squashed anyway when
they are merged.

lysnikolaou:
Got it!

pegen/pegen.c
gvanrossum: I find the current `seq_to_dotted_name()` implementation too fragile, so I'd rather make that change now.

pegen/pegen.c
lysnikolaou: @gvanrossum I still use `PyUnicode_GET_LENGTH` here, since `PyUnicode_GET_SIZE` is deprecated since 3.3 and I could not find something better. What would be the correct function to use here?

data/simpy.gram
```diff
@@ -53,7 +53,9 @@ dotted_as_names[asdl_seq*]: a=dotted_as_name b=(',' c=dotted_as_name { c })* { s
 dotted_as_name[alias_ty]:
     | a=dotted_name 'as' b=NAME { _Py_alias(((expr_ty) a)->v.Name.id, ((expr_ty) b)->v.Name.id, p->arena) }
     | a=dotted_name { _Py_alias(((expr_ty) a)->v.Name.id, NULL, p->arena) }
-dotted_name[expr_ty]: a=NAME b=('.' c=NAME { c })* { seq_to_dotted_name(p, seq_insert_in_front(p, a, b)) }
+dotted_name[expr_ty]:
+    | a=NAME b=('.' c=dotted_name { c }) { join_names_with_dot(p, a, b) }
```
gvanrossum: This can be simplified to `a=NAME '.' b=dotted_name { join_names_with_dot(p, a, b) }`, i.e. without the parentheses.

Also it would seem to better match the semantics to make it left-recursive. (But right-recursive is probably more efficient.)

pegen/pegen.c
```diff
@@ -505,22 +505,22 @@ seq_flatten(Parser *p, asdl_seq *seqs)
...
-seq_to_dotted_name(Parser *p, asdl_seq *seq)
+join_names_with_dot(Parser *p, expr_ty first_name, expr_ty second_name)
 {
-    /* Create a string of the form a.b.c like Python/ast.c alias_for_dotted_name does
-       for dotted names */
-    if (asdl_seq_LEN(seq) == 1) {
-        return asdl_seq_GET(seq, 0);
-    }
+    /* Create a string of the form "first_name.second_name" like
+       Python/ast.c alias_for_dotted_name does for dotted names */
```
gvanrossum: This comment doesn't add a lot of value (and eventually there won't be a Python/ast.c :-).

pegen/pegen.c
```diff
@@ -505,22 +505,22 @@ seq_flatten(Parser *p, asdl_seq *seqs)
...
-        len += current_len + 1;
+    if (PyUnicode_READY(first_identifier) == -1) {
+        return NULL;
+    }
+    if (PyUnicode_READY(second_identifier) == -1) {
+        return NULL;
     }
-    --len; // Last name does not have a dot
+    ssize_t len = PyUnicode_GET_LENGTH(first_name->v.Name.id) +
+                  PyUnicode_GET_LENGTH(second_name->v.Name.id) + 1; // +1 for the dot
```
gvanrossum: You should really calculate the length using `strlen(first_str) + 1 + strlen(second_str)`, then you get it right. And move the code around of course.

lysnikolaou:
Thanks for the (many) quick reviews!

Issue #72, opened 2019-11-11
Problem (?) with EXTRA in import_name and import_from


lysnikolaou:
Going through all the things I did for #67 and #71, I realised that there might be a slight problem we didn't catch, but I am not sure.

Let me explain with the example of the `import_name` rule:
```
import_name[stmt_ty]: a='import' b=dotted_as_names { _Py_Import(b, EXTRA(a, b)) }
```

`b` is an `asdl_seq *` and we use it as the tail parameter to `EXTRA`, which calls `ENDLINE` and `ENDCOL` for `b`, which then access the `end_lineno` and `end_col_offset` attributes after casting the `asdl_seq *` to an `expr_ty`. I don't really see how that doesn't generate a SEGFAULT, but (SEGFAULTS aside) I think it will probably be a bug in the future, when these AST nodes are used for error reporting.

Am I maybe missing something?

gvanrossum:
Ouch, definitely a real problem. Good catch! Not sure what to do, maybe a
helper function?
-- 
--Guido (mobile)

lysnikolaou:
I think that a helper function is the way to go as well. Something like `asdl_seq_get_tail_elem()`?

gvanrossum:
Yeah, though what to do if the length is zero? Maybe pass the previous item
in as well?

We should also try to make it more likely that this is caught by the C type
checker.
-- 
--Guido (mobile)

lysnikolaou:
> We should also try to make it more likely that this is caught by the C type checker.

The problem with that is that the we need the cast to `expr_ty`, which makes it possible to pass in whatever pointer type we want and the C type checker will never say anything about it.

gvanrossum:
Huh, does only expr_ty have line/column info?

lysnikolaou:
No, but, to the best of my knowledge, the only general types that do are `expr_ty` and `stmt_ty` (there is also `arg_ty` and `excepthandler_ty`, but those aren't general enough for us to cast to). The problem is that an `asdl_seq *` can/will hold `void *`s, which means that we _need_ to cast to either of those, in order to be able to access line/column info (I think it doesn't really matter to which exactly). Due to this we can't really get much help from the type checker.

We could cast in the helper function though, where we know if the element comes from an `asdl_seq *` or not, but the type checker still doesn't come into play.

gvanrossum:
Hm, yeah, C doesn't have generics... :-(

I guess we'll have to just live dangerously and document what we're doing.

[...]

PR #73, opened 2019-11-12
Fix EXTRA calls (Correct arguments and correct casting)

lysnikolaou:
After all, there were more problems that I had foreseen, so I had to delay my PR.

The last problem I was able to recognize, was that `alias_ty` does not contain any line/col info. In Python/ast.c the respective function only relies on the node info to extract line and col. To remedy this, I created a wrapper type `PegenAlias` which contains all the useful info and gets it from the `expr_ty`s containing the names, with which the aliases get created.

It still doesn't work 100%, but it is a good start. There is a bug with line/col of the outermost statement not getting the correct value, which does not occur during debugging, so it might take some time to figure out where it comes from. Any ideas maybe?

Fix for #72.


data/simpy.gram
```diff
@@ -20,7 +20,7 @@ small_stmt[stmt_ty]:
     | break_stmt
     | continue_stmt
     | assignment
-    | e=expressions { _Py_Expr(e, EXTRA(e, e)) }
+    | e=expressions { _Py_Expr(e, EXTRA(e, expr_type, e, expr_type)) }
```
gvanrossum: Why not `EXTRA_EXPR(e, e)`?

pegen/pegen.c
```diff
@@ -592,3 +592,90 @@ alias_for_star(Parser *p)
...
+    asdl_seq *new_seq = _Py_asdl_seq_new(len, p->arena);
+    for (int i = 0; i < len; i++) {
+        PegenAlias *a = asdl_seq_GET(seq, i);
+        asdl_seq_SET(new_seq, i, a->alias);
+    }
+    return new_seq;
+}
+
+int
+expr_type(void *a, int head, int line)
```
gvanrossum: Why not change the type of `a` to `expr_ty`? If the caller passes a `void *` it'll still pass the compiler, and if something else is passed, it's probably an error. Then you won't need the casts in the function body. (Ditto below.)

pegen/pegen.h
```diff
@@ -49,11 +54,12 @@ void *keyword_token(Parser *p, const char *val);
 
 void *CONSTRUCTOR(Parser *p, ...);
 
-#define LINE(arg) ((expr_ty)(arg))->lineno
-#define COL(arg) ((expr_ty)(arg))->col_offset
-#define ENDLINE(arg) ((expr_ty)(arg))->end_lineno
-#define ENDCOL(arg) ((expr_ty)(arg))->end_col_offset
-#define EXTRA(head, tail) LINE(head), COL(head), ENDLINE(tail), ENDCOL(tail), p->arena
+#define EXTRA_EXPR(head, tail) EXTRA(head, expr_type, tail, expr_type)
+#define EXTRA(head, head_type_func, tail, tail_type_func) head_type_func(head, 1, 1), \
```
gvanrossum: It seems a little perverse that the fundamental building block has the shorter name while the convenience macros have a longer name.

test/test_ast_generation.py
```diff
@@ -26,15 +28,15 @@ def read_python_source(path: str) -> str:
     return source
 
 
-def test_ast_generation_on_source_files(tmp_path: PurePath) -> None:
+@pytest.mark.parametrize("filename", PYTHON_SOURCE_FILENAMES)
```
gvanrossum: Thanks for the refactoring! However now we build the extension for each file, which is rather slow. Maybe we can use pytest's fixture machinery to create the extension once and pass it in to each test?

test/test_ast_generation.py
```diff
@@ -26,15 +28,15 @@ def read_python_source(path: str) -> str:
...
-    for filename in PYTHON_SOURCE_FILENAMES:
-        print(filename)
-        source = read_python_source(os.path.join(TEST_DIR, filename))
-
-        actual_ast = extension.parse_string(source)
-        expected_ast = ast.parse(source)
-        assert ast.dump(actual_ast) == ast.dump(
-            expected_ast
-        ), f"Wrong AST generation for file: {filename}"
+    print(filename)
```
gvanrossum: You don't need to print this any more, since it's in the assert message.

data/simpy.gram
```diff
@@ -36,44 +36,50 @@ assert_stmt: 'assert' expression [',' expression]
...
 import_stmt: import_name | import_from
-import_name[stmt_ty]: a='import' b=dotted_as_names { _Py_Import(b, EXTRA(a, b)) }
+import_name[stmt_ty]: a='import' b=dotted_as_names {
+    _Py_Import(seq_map_to_alias(p, b), EXTRA(a, token_type, seq_get_tail(NULL, b), alias_type)) }
 # note below: the ('.' | '...') is necessary because '...' is tokenized as ELLIPSIS
 import_from[stmt_ty]:
     | a='from' b=('.' | '...')* !'import' c=dotted_name 'import' d=import_from_targets {
-        _Py_ImportFrom(((expr_ty) c)->v.Name.id, d, seq_count_dots(b), EXTRA(a, d)) }
-    | e='from' f=('.' | '...')+ 'import' g=import_from_targets { _Py_ImportFrom(NULL, g, seq_count_dots(f), EXTRA(e, g)) }
+        _Py_ImportFrom(((expr_ty) c)->v.Name.id, seq_map_to_alias(p, d), seq_count_dots(b), EXTRA(a, token_type, seq_get_tail(NULL, d), alias_type)) }
```
gvanrossum: Do you still need the cast? I'd think not, `dotted_name` is typed correctly. Also a few more times below I believe.

Separately, maybe break up the long line? A line is too long when GitHub starts using a horizontal scroll bar. :-)

data/simpy.gram
```diff
@@ -36,44 +36,50 @@ assert_stmt: 'assert' expression [',' expression]
...
-    | 'if' a=full_expression ':' b=block c=else_block { _Py_If(a, b, c, EXTRA(a, c)) }
-    | 'if' a=full_expression ':' b=block c=elif_stmt { _Py_If(a, b, singleton_seq(p, c), EXTRA(a, c)) }
-    | 'if' a=full_expression ':' b=block { _Py_If(a, b, NULL, EXTRA(a, b)) }
-elif_stmt:
-    | 'elif' a=full_expression ':' b=block c=else_block { _Py_If(a, b, c, EXTRA(a, c)) }
-    | 'elif' a=full_expression ':' b=block c=elif_stmt { _Py_If(a, b, singleton_seq(p, c), EXTRA(a, c)) }
-    | 'elif' a=full_expression ':' b=block { _Py_If(a, b, NULL, EXTRA(a, b)) }
-else_block: 'else' ':' b=block { b }
+if_stmt[stmt_ty]:
+    | a='if' b=full_expression ':' c=block d=else_block { _Py_If(b, c, d, EXTRA(a, token_type, seq_get_tail(NULL, d), stmt_type)) }
```
gvanrossum: I think it's time to name the variables more meaningfully than `a`, `b`, `c` and `d`. (Note: never name a variable `p`. :-)

pegen/pegen.h
```diff
@@ -49,11 +54,12 @@ void *keyword_token(Parser *p, const char *val);
...
+                                                          head_type_func(head, 1, 0), \
+                                                          tail_type_func(tail, 0, 1), \
+                                                          tail_type_func(tail, 0, 0), \
+                                                          p->arena
```
gvanrossum: I have another idea. Maybe we can use C preprocessor "token pasting" (`token1##token2`) instead of the helper signature that takes `head, line` args. So we could do something like
```
...
head_type_func##headline(head), \
head_type_func##headcol(head), \
tail_type_func##tailline(tail), \
tail_type_func##tailcol(tail), \
...
```
And the definitions would be e.g.
```
int expr_type_headline(expr_ty a) { return a->line; }
int expr_type_headcol(expr_ty a) { return a->col_offset; }
```
etc.

Or actually they could even be macros:
```
#define expr_type_headline(a) (a->line)
```
etc.

pegen/pegen.h
```diff
@@ -26,6 +26,11 @@ typedef struct {
     PyArena *arena;
 } Parser;
 
+typedef struct {
+    alias_ty alias;
+    int line, col, endline, endcol;
```
gvanrossum: Honestly I still think there will be some utility in naming these variables the same as they are in `expr_ty` (even if their offsets aren't). And ditto for Token.

data/simpy.gram
lysnikolaou: Just an oversight from my side, I guess.

test/test_ast_generation.py
lysnikolaou: That's probably the best way to go, yes!

pegen/pegen.h
lysnikolaou: @gvanrossum That indeed looks like a better idea! I'll implement it ASAP and the start testing to find out what goes wrong with the line/col info.

Is there maybe a way to globally disable memoization? In the rules I want to debug, it is not really needed and it makes debugging a bit more difficult for me, since I haven't still understood it 100%.

gvanrossum: You can disable memoization in the generator. At the top of visit_Rule() there's a line `memoize = not node.left_recursive`. Just set that to False and you should be good. However for left-recursive rules you'll still get use of the memo, the left-recursion algorithm depends on it. (But it doesn't use that flag, so just setting `memoize = False` should work.)

pegen/pegen.h
```diff
@@ -54,11 +54,31 @@ void *keyword_token(Parser *p, const char *val);
 
 void *CONSTRUCTOR(Parser *p, ...);
 
+#define expr_type_headline(a) ((expr_ty) a)->lineno
```
gvanrossum: I wonder if these should be made real functions after all. We can then make the argument types `expr_ty` etc. which would catch mistakes where something that's known to be e.g. `stmt_ty` will be flagged as an error. `void *` will still work. The functions can be made `inline` for performance.

lysnikolaou: Done! I was not sure, if it's okay to put these directly in the header file, so let me know if it is.

pegen/pegen.c
```diff
@@ -613,7 +613,7 @@ pegen_alias(alias_ty alias, int line, int endline, int col, int endcol, PyArena
     a->lineno = line;
     a->end_lineno = endline;
     a->col_offset = col;
```
gvanrossum: I'd swap the above two lines, just to keep things in a consistent order with the arguments. Maybe also make sure that the fields are declared in the same order -- it should always be `lineno, col_offset, end_lineno, end_col_offset`. Making the names consistent everywhere also might be a good idea -- just so there's less need to think. :-)

test/test_ast_generation.py
```diff
@@ -3,6 +3,8 @@
 from pathlib import PurePath
 from typing import Any
 
+import pytest
```
gvanrossum: I think you need to add `# type: ignore` to this line to make mypy happy. (Unfortunately the pytest API has many items and there are no mypy stubs yet.)

lysnikolaou:
@gvanrossum After further inspection with gdb, it seems that most errors in `import_*.py` files, when run under a normal interactive interpreter session and not with `pytest`, only occur after `parse.parse_file(...)` has been called at least once. Could it be, that some state remains after the first call to the function, which leads to errors in successive calls?

gvanrossum:
> @gvanrossum After further inspection with gdb, it seems that most errors in `import_*.py` files, when run under a normal interactive interpreter session and not with `pytest`, only occur after `parse.parse_file(...)` has been called at least once. Could it be, that some state remains after the first call to the function, which leads to errors in successive calls?

That's possible, but it could also be a reference count error. It's also possible that an exception is set but ignored (e.g. by some code not checking for a NULL result from a function that may set an exception) and then found out much later when some other function *does* check for NULL.

You should probably figure out how to build a debug version of Python 3.8 and use that binary -- it contains asserts that catch many of these issues earlier.

lysnikolaou:
> You should probably figure out how to build a debug version of Python 3.8 and use that binary -- it contains asserts that catch many of these issues earlier.

Ok, I'll do that then.

lysnikolaou:
I can't really understand why valgrind gives me this error, but it says that the keyword token `a` in `if_stmt_rule`, which we are trying to call `token_type_headcol` with, has been free`d due to the realloc [here](https://github.com/gvanrossum/pegen/blob/1c93b8070875bd2da7519f1aa4fd2f0c74121f50/pegen/pegen.c#L78).

Here's the full log:
```
>>> t = parse.parse_file('test/test_data/if_else.py')
==30187== Invalid read of size 4
==30187==    at 0x54BDA3D: token_type_headcol (pegen.h:85)
==30187==    by 0x54BDA3D: if_stmt_rule (parse.c:1469)
==30187==    by 0x54BE281: compound_stmt_rule (parse.c:712)
==30187==    by 0x54BE307: statement_rule (parse.c:521)
==30187==    by 0x54BE36C: _loop1_70_rule (parse.c:6792)
==30187==    by 0x54BE4B1: statements_rule (parse.c:500)
==30187==    by 0x54BE4E5: start_rule (parse.c:478)
==30187==    by 0x54B4716: run_parser (pegen.c:349)
==30187==    by 0x54B491C: run_parser_from_file (pegen.c:410)
==30187==    by 0x54BE558: parse_file (parse.c:8177)
==30187==    by 0x1776C6: cfunction_call_varargs (call.c:757)
==30187==    by 0x17A83B: PyCFunction_Call (call.c:772)
==30187==    by 0x1782CA: _PyObject_MakeTpCall (call.c:159)
==30187==  Address 0x538bbe4 is 36 bytes inside a block of size 64 free'd
==30187==    at 0x483CFAF: realloc (in /usr/lib/x86_64-linux-gnu/valgrind/vgpreload_memcheck-amd64-linux.so)
==30187==    by 0x1B82C7: _PyMem_RawRealloc (obmalloc.c:121)
==30187==    by 0x1B8C1E: _PyMem_DebugRawRealloc (obmalloc.c:2254)
==30187==    by 0x1B8E4C: _PyMem_DebugRealloc (obmalloc.c:2339)
==30187==    by 0x1B9807: PyMem_Realloc (obmalloc.c:623)
==30187==    by 0x54B3CE5: fill_token (pegen.c:78)
==30187==    by 0x54B40C9: expect_token (pegen.c:168)
==30187==    by 0x54B4149: name_token (pegen.c:203)
==30187==    by 0x54B7BE6: full_expression_rule (parse.c:2444)
==30187==    by 0x54BD8D6: if_stmt_rule (parse.c:1461)
==30187==    by 0x54BE281: compound_stmt_rule (parse.c:712)
==30187==    by 0x54BE307: statement_rule (parse.c:521)
==30187==  Block was alloc'd at
==30187==    at 0x483A7F3: malloc (in /usr/lib/x86_64-linux-gnu/valgrind/vgpreload_memcheck-amd64-linux.so)
==30187==    by 0x1B8318: _PyMem_RawMalloc (obmalloc.c:99)
==30187==    by 0x1B88CF: _PyMem_DebugRawAlloc (obmalloc.c:2134)
==30187==    by 0x1B890B: _PyMem_DebugRawMalloc (obmalloc.c:2167)
==30187==    by 0x1B8930: _PyMem_DebugMalloc (obmalloc.c:2316)
==30187==    by 0x1B979A: PyMem_Malloc (obmalloc.c:605)
==30187==    by 0x54B469E: run_parser (pegen.c:328)
==30187==    by 0x54B491C: run_parser_from_file (pegen.c:410)
==30187==    by 0x54BE558: parse_file (parse.c:8177)
==30187==    by 0x1776C6: cfunction_call_varargs (call.c:757)
==30187==    by 0x17A83B: PyCFunction_Call (call.c:772)
==30187==    by 0x1782CA: _PyObject_MakeTpCall (call.c:159)
```

lysnikolaou:
Oh, I think I've got it! A `void *` gets returned by `keyword_token` that at the time holds the token for the keyword `if`, but after that the `Parser.tokens` gets reallocated, which means that the previous address of `tokens` gets free'd, which leads to a use-after-free error.

lysnikolaou:
> Does this mean that every token now gets copied? That leads to a slightly different solution: make `p->tokens` an array of `Token *`. It also avoids the leak you have now introduced (the copy _never_ gets freed).

I was maybe a bit too enthusiastic and missed that. `p->tokens` is now an array of `Token *`s. 

I also broke up long lines in the Grammar trying to follow a somewhat consistent style. Let me know if anything doesn't seem okay.

lysnikolaou:
Yes! This certainly wasn't an easy one!

PR #75, opened 2019-11-20
Grammar refactorings for expressions

lysnikolaou:
* Add actions to BoolOp and BinOp rules
* Add CmpopExprPair, which stores the necessary info for comparison

Points that may need further discussion:
1. Is there a better way to propagate (cmpop, comparator) pairs to the comparison rule?
2. Is my understanding of the `expressions` rule correct in that it returns a tuple in case there's a comma? (i.e Is the split to all the different alternatives correct/necessary?)
3. I had a look at Python/ast.c to see what it does about the STRING+ thing and it basically just concatenates strings together, which is far too complex as it needs to handle all the different kind of strings. Should I go ahead and start implementing something similar here or do we have a better way of handling it?


gvanrossum:
Hi, I am super busy today and away from my laptop. Sorry!

lysnikolaou:
No worries at all! I just had some time to spare and decided to make a PR.

lysnikolaou:
Thanks!

PR #76, opened 2019-11-25
Implement dot operator

lysnikolaou:
Implement dot operator, which gathers the parsed nodes into a sequence, ignoring their separators(such as a comma):
    * Add node in grammar.py
    * Generate correct Python code for new node

I didn't get to implement the generation of C code for this, but I'd like to have a 'go' or 'no go' before I do, regarding the way I implemented this. Could it be maybe done better?

### DIFF
--- a/pegen/grammar.py
+++ b/pegen/grammar.py
@@ -72,7 +72,7 @@ class Rule:
     def is_loop(self) -> bool:
         return self.name.startswith("_loop")
 
-    def is_tmp_with_sep(self) -> bool:
+    def is_node_with_sep(self) -> bool:
         return self.name.startswith("_tmp_sep")
 
     def __str__(self) -> str:
--- a/pegen/grammar.py
+++ b/pegen/grammar.py
@@ -72,8 +72,8 @@ class Rule:
     def is_loop(self) -> bool:
         return self.name.startswith("_loop")
 
-    def is_loop_with_sep(self) -> bool:
-        return self.name.startswith("_loop_sep")
+    def is_tmp_with_sep(self) -> bool:
+        return self.name.startswith("_tmp_sep")
 
     def __str__(self) -> str:
         if self.type is None:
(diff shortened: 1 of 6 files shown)


pegen/python_generator.py
```diff
@@ -107,6 +108,13 @@ def visit_Repeat1(self, node: Repeat1) -> Tuple[str, str]:
         self.cache[node] = name, f"self.{name}()"  # But no trailing comma here!
         return self.cache[node]
 
+    def visit_RepeatWithSeparator(self, node: RepeatWithSeparator) -> Tuple[str, str]:
+        if node in self.cache:
+            return self.cache[node]
+        name = self.gen.name_node_with_sep(node)
+        self.cache[node] = name, f"self.{name}()"  # But no trailing comma here!
```
gvanrossum: I'd change this comment to "# No trailing comma here either!" (so the sequence of comments in these 4 methods reads somewhat humorously).

pegen/grammar.py
```diff
@@ -72,6 +72,9 @@ def __init__(self, name: str, type: Optional[str], rhs: Rhs):
     def is_loop(self) -> bool:
         return self.name.startswith("_loop")
 
+    def is_node_with_sep(self) -> bool:
```
gvanrossum: I'm not excited about the long name here (esp. since it's repeated in a keyword arg later).

gvanrossum:
I guess now we can start using it...

lysnikolaou:
I am currently working on implementing the C code generation for this. I thought that this would go into this PR. Should I create a new one?

gvanrossum:
Oops. I guess we need a convention: if you put [WIP] in the subject of a PR
that means not to merge it yet even if it's perfect. Go ahead and make a
new PR here (I like smaller steps anyway, but I like squashing PRs since
the local commits are rarely of historical value later).

lysnikolaou:
I'll do that in the future then.

PR #77, opened 2019-11-25
Grammar refactorings for try_stmt

lysnikolaou:
* Add actions to try, except and finally rules
* Dummy action in target rule for now

### DIFF
--- a/data/simpy.gram
+++ b/data/simpy.gram
@@ -106,10 +106,25 @@ for_stmt: [ASYNC] 'for' star_targets 'in' expressions ':' block [else_block]
 
 with_stmt: [ASYNC] 'with' expression ['as' target] (',' [expression ['as' target]])* ':' block
 
-try_stmt: try_block finally_block | try_block except_block+ [else_block] [finally_block]
-try_block: 'try' ':' block
-except_block: 'except' [expression ['as' target]] ':' block
-finally_block: 'finally' ':' block
+try_stmt[stmt_ty]:
+    | a='try' ':' b=block ex=except_block+ el=else_block f=finally_block {
+        _Py_Try(b, ex, el, f, EXTRA(a, token_type, seq_get_tail(NULL, f), stmt_type)) }
+    | a='try' ':' b=block ex=except_block+ el=else_block {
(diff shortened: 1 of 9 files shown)


data/simpy.gram
```diff
@@ -106,10 +106,25 @@ for_stmt: [ASYNC] 'for' star_targets 'in' expressions ':' block [else_block]
...
+        _Py_Try(b, ex, NULL, NULL, EXTRA(a, token_type, seq_get_tail(NULL, ex), excepthandler_type)) }
+    | a='try' ':' b=block f=finally_block {
+        _Py_Try(b, NULL, NULL, f, EXTRA(a, token_type, seq_get_tail(NULL, f), stmt_type)) }
+except_block[excepthandler_ty]:
+    | a='except' e=expression 'as' t=target ':' b=block {
+        _Py_ExceptHandler(e, t->v.Name.id, b, EXTRA(a, token_type, seq_get_tail(NULL, b), stmt_type)) }
+    | a='except' e=expression ':' b=block {
+        _Py_ExceptHandler(e, NULL, b, EXTRA(a, token_type, seq_get_tail(NULL, b), stmt_type)) }
+    | a='except' ':' b=block {
+        _Py_ExceptHandler(NULL, NULL, b, EXTRA(a, token_type, seq_get_tail(NULL, b), stmt_type)) }
```
gvanrossum: The expansion of optional parts here makes me sad. Doesn't an optional item return `NULL` if not found? (I may actually be wrong about this -- it's been a while since I wrote this code and right now I am tired and jet-lagged. :-)

lysnikolaou: That is right. The problem with that approach is that we will have a deeply nested call to `seq_get_tail()`, which doesn't make this a lot clearer.

It'd look like this:
```
    a='try' ':' b=block ex=except_block+ el=[else_block] f=[finally_block]
        _Py_Try(b, ex, el, f, EXTRA(a, token_type, seq_get_tail(seq_get_tail(seq_get_tail(NULL, seq_get_tail(NULL, ex)), el), f), stmt_type))
```
which is unworkable I think.

gvanrossum: Varargs functions in C are a pain, so let's stick to (1) for now. But let's keep this in mind for the future, maybe we'll think of a better way.

PR #78, opened 2019-11-28
Implementation of C Code Generation for the dot Operator

### LINKED ISSUE #70: Add custom rule for repetitions separated by another token
As noted in #67 the construct `X (sep X)*` gets heavily used by the current grammar, at least 16 times according to a quick search on my editor. It would be a good idea to add a custom rule to handle this construct, so that we don't have to use `seq_insert_in_front` every time, as it creates a brand new `asdl_seq*`. What would the best design be for such a rule?

lysnikolaou:
I'm not sure if we want to use `seq_insert_in_front` for this, but this seemed easier, so we can start with it and optimize later.

I also renamed `RepeatWithSeparator` to `Gather`, which seems better.

Closes #70.

### DIFF
--- a/pegen/c_generator.py
+++ b/pegen/c_generator.py
@@ -17,6 +17,7 @@ from pegen.grammar import (
     Opt,
     Repeat0,
     Repeat1,
+    Gather,
     Group,
     Rule,
 )
@@ -155,6 +156,13 @@ class CCallMakerVisitor(GrammarVisitor):
         self.cache[node] = f"{name}_var", f"{name}_rule(p)"  # But not here!
         return self.cache[node]
 
+    def visit_Gather(self, node: Gather) -> Tuple[str, str]:
+        if node in self.cache:
+            return self.cache[node]
+        name = self.gen.name_gather(node)
+        self.cache[node] = f"{name}_var", f"{name}_rule(p)"  # No trailing comma here either!
+        return self.cache[node]
+
     def visit_Group(self, node: Group) -> Tuple[Optional[str], str]:
         return self.visit(node.rhs)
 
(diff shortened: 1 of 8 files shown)


pegen/c_generator.py
```diff
@@ -155,6 +156,13 @@ def visit_Repeat1(self, node: Repeat1) -> Tuple[str, str]:
         self.cache[node] = f"{name}_var", f"{name}_rule(p)"  # But not here!
         return self.cache[node]
 
+    def visit_Gather(self, node: Gather) -> Tuple[str, str]:
+        if node in self.cache:
+            return self.cache[node]
+        name = self.gen.name_gather(node)
+        self.cache[node] = f"{name}_var", f"{name}_rule(p)"  # No trailing comma here either!
```
gvanrossum: I'd just drop the comment here and on line 156 above; it no longer feels funny, and in C it's not about a trailing comma but about a comma operator.

pegen/grammar.py
```diff
@@ -382,7 +382,7 @@ def nullable_visit(self, rules: Dict[str, Rule]) -> bool:
         return False
 
 
-class RepeatWithSeparator(Repeat):
+class Gather(Repeat):
```
gvanrossum: Note: a few lines below, the `__str__()` method wasn't changed when you required the trailing `+` per @apalala's suggestion.

pegen/grammar.py
```diff
@@ -73,7 +73,7 @@ def is_loop(self) -> bool:
         return self.name.startswith("_loop")
 
     def is_gather(self) -> bool:
-        return self.name.startswith("_tmp_sep")
+        return self.name.startswith("_tmp_gather")
```
gvanrossum: Let's drop the `_tmp` prefix -- the rules for loops don't use this either.

PR #87, opened 2019-12-03
Refactorings for all small stmt rules (but assignment)

lysnikolaou:
* return_stmt
* yield_stmt
* raise_stmt
* assert_stmt
* del_stmt
* global_stmt
* nonlocal_stmt


data/simpy.gram
```diff
@@ -183,6 +193,7 @@ full_expression: NAME ':=' expression | expression
 yield_expression: yield_expr | expression
 expression[expr_ty]:
     | lambdef
+    | yield_expr
```
gvanrossum: This doesn't look right, `yield_expr` is also in `yield_expression` one line above. Maybe that `yield_stmt` existed for a reason? Also a lambda expression is *not* allowed to contain an unparenthesized yield expression.

data/simpy.gram
```diff
@@ -326,7 +340,7 @@ star_targets: a=star_target (',' !'in' star_target)* [','] { a }
 star_target: '*' bitwise_or | atom t_tail+ | a=star_atom t_tail* { a }
 star_atom: a=NAME { store_name(p, a) } | '(' [star_targets] ')' | '[' [star_targets] ']'
 
-targets: target (',' target)* [',']
+targets[asdl_seq*]: a=','.target+ [','] { a }
```
gvanrossum: I predict a merge conflict on this line. :-)

pegen/pegen.c
```diff
@@ -641,6 +641,17 @@ asdl_seq *seq_map_to_alias(Parser *p, asdl_seq *seq)
     return new_seq;
 }
 
+asdl_seq *names_map_to_id(Parser *p, asdl_seq *seq)
```
gvanrossum: I'm not sure I understand how you came with this name. I'd put the verb first and keep the plurality the same, e.g. 'map_names_to_ids'.

Also the type should go on a line by itself.

pegen/pegen.c
```diff
@@ -693,3 +704,27 @@ store_name(Parser *p, expr_ty load_name)
...
+expr_ty
+del_name(Parser *p, expr_ty load_name)
+{
+    return _Py_Name(load_name->v.Name.id,
+                    Del,
+                    EXTRA_EXPR(load_name, load_name));
+}
+
+asdl_seq *
+targets_map_to_del_name(Parser *p, asdl_seq *seq)
```
gvanrossum: Same comment on naming here.

Also I wonder if we should start adding a brief block comment to each helper function explaining what it does. (Then again maybe just changing the name is sufficient for now.)

pegen/pegen.c
```diff
@@ -693,3 +704,27 @@ store_name(Parser *p, expr_ty load_name)
...
+{
+    int len = asdl_seq_LEN(seq);
+    asdl_seq *new_seq = _Py_asdl_seq_new(len, p->arena);
+    for (int i = 0; i < len; i++) {
+        expr_ty e = asdl_seq_GET(seq, i);
+        if (e->kind == Name_kind) {
```
gvanrossum: Why is this check needed? This makes my spider sense tickle...

test/test_data/delete.py
```diff
@@ -0,0 +1 @@
+del a
```
gvanrossum: I see no test for `del a, b` -- not to speak of `del a, (b, c)`.

data/simpy.gram
lysnikolaou: I have resolved this conflict locally, would it be okay to force-push this or should I do something else?

pegen/pegen.c
lysnikolaou: Well, I just followed the naming I had introduced with `seq_map_to_alias`. I'll change all three of them.

data/simpy.gram
lysnikolaou: I thought that this would be okay, since `simpy_cpython` still succeeds. Do you think this would be better?

```
small_stmt[stmt_ty]:
    | yield_stmt

yield_stmt[stmt_ty]: a=yield_expr { _Py_Expr(a, EXTRA_EXPR(a, a)) }
```

gvanrossum: Yeah, that should work.

The reason `make simpy_cpython` doesn't fail is that there are no negative tests. Try this:
```
lambda: yield
```
That is an error in regular Python, but the change you propose in this PR would make it pass. (To get the same effect in regular Python, use `lambda: (yield)`.)

data/simpy.gram
gvanrossum: Typically one would use `git merge` but if you've already rebased I am okay with a force push.

pegen/pegen.c
gvanrossum: I guess I didn't object to `seq_map_to_alias` because you have a bunch of other `seq_*` functions. But now I look at it, it seems that that one should really be called `map_pegen_aliases_to_aliases`, which is a bit long... Maybe that one should be `extract_orig_aliases`? Because it seems it takes a sequence of PegenAliases (which wrap an alias_ty together with line/col info) and returns a sequence of those alias_ty values.

data/simpy.gram
lysnikolaou: I'll keep that in mind for the future.

data/simpy.gram
lysnikolaou: Got it.

pegen/pegen.c
lysnikolaou: Done.

lysnikolaou:
@gvanrossum I just force-pushed, after I rebased master and also committed some new things. One things I did is change `map_targets_to_del_names` to correctly handle tuples and lists (or even nested tuples). That was what I understood after going through the targets rules, but I may be wrong. Are these the only cases or should something else be covered as well?

pegen/pegen.c
```diff
@@ -693,3 +706,33 @@ store_name(Parser *p, expr_ty load_name)
...
+            asdl_seq_SET(new_seq, i, del_name(p, e));
+        } else if (e->kind == Tuple_kind) {
+            asdl_seq_SET(new_seq, i, _Py_Tuple(map_targets_to_del_names(p, e->v.Tuple.elts),
+                                               Del,
+                                               EXTRA_EXPR(e, e)));
+        } else if (e->kind == List_kind) {
+            asdl_seq_SET(new_seq, i, _Py_List(map_targets_to_del_names(p, e->v.List.elts),
+                                              Del,
+                                              EXTRA_EXPR(e, e)));
+        }
```
gvanrossum: There are other possible kinds here -- a target may also take the form `a[i]` or `o.a`. Since those aren't yet supported in the grammar (they have no actions) it doesn't matter, but I'd at least add a comment, or if it doesn't break the tests an assert.

lysnikolaou:
Thanks for the review and sorry that this was a bit messy at the beginning.

PR #92, opened 2019-12-05
Refactor rules for with_stmt

lysnikolaou:


lysnikolaou:
I like that! I'll add it.

test/test_c_parser.py
```diff
@@ -227,3 +227,29 @@ def test_same_name_different_types(tmp_path: PurePath) -> None:
...
+        e=NAME o=['as' t=NAME { t }] { _Py_withitem(e, store_name(p, o), p->arena) }
+    )
+    block[stmt_ty]: a=pass_stmt NEWLINE { a } | NEWLINE INDENT a=pass_stmt DEDENT { a }
+    pass_stmt[stmt_ty]: a='pass' { _Py_Pass(EXTRA(a, token_type, a, token_type)) }
+    """
+    stmt = "with (\n    a as b,\n    c as d\n): pass"
+    grammar = parse_string(grammar_source, GrammarParser)
+    extension = generate_parser_c_extension(grammar, tmp_path)
+    the_ast = extension.parse_string(stmt)
+    assert ast.dump(the_ast).startswith(
```
gvanrossum: Why starts with and not ==?

lysnikolaou: So that I don't have to include unnecessary noise in the string such as `body=([Pass()])`.

PR #93, opened 2019-12-06
Documentation about helper functions and grammar syntax

lysnikolaou:
I have written down some things about the helper functions and the grammar syntax. It's just a few words for each. 

Beware that my wording might not be the best in some cases and I may have missed some things.

### DIFF
--- a/docs/grammar.md
+++ b/docs/grammar.md
@@ -1,0 +1,10 @@
+Style
+-----
+This is not a hard limit, but lines longer than 100 characters should be
+wrapped.  Most lines should be wrapped after the opening action curly brace,
+like:
+
+```
+really_long_rule[expr_ty]: some_arbitrary_rule {
+    _This_is_the_action }
+```
--- a/docs/grammar.md
+++ b/docs/grammar.md
@@ -3,17 +3,17 @@ Syntax
 
 The grammar consists of a sequence of rules of the form:
 ```
-rule_name: <expression>
+rule_name: expression
 ```
 
 Optionally, a type can be included right after the rule name,
-which specifies the return type of the C function corresponging
-to the rule:
+which specifies the return type of the C or Python function
+corresponding to the rule:
 ```
-rule_name[return_type]: <expression>
+rule_name[return_type]: expression
 ```
 If the return type is omitted, then a `void *` is
-returned.
+returned in C and an `Any` in Python.
 
 ### Grammar Expressions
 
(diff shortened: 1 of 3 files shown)


docs/grammar.md
```diff
@@ -0,0 +1,90 @@
...
+Python-style comments.
+
+##### `e1 e2`
+Match e1, then match e2.
+
+##### `e1 | e2`
+Match e1 or e2.
+
+A | can be used before the first alternative as well for formatting
+purposes.
```
gvanrossum: This only applies when the alternatives are formatted on separate lines (like in the return value example below), not in the "short form":
```
rule_name: | alternative1 | alternative2
```
is invalid.

docs/grammar.md
```diff
@@ -0,0 +1,90 @@
+Syntax
+------
+
+The grammar consists of a sequence of rules of the form:
+```
+rule_name: <expression>
+```
+
+Optionally, a type can be included right after the rule name,
+which specifies the return type of the C function corresponging
```
gvanrossum: It's not just C functions. Types are supported when generating Python code too.

docs/grammar.md
```diff
@@ -0,0 +1,90 @@
...
+to the rule:
+```
+rule_name[return_type]: <expression>
+```
+If the return type is omitted, then a `void *` is
+returned.
```
gvanrossum: When generating Python, the default type is `Any`.

docs/grammar.md
```diff
@@ -0,0 +1,90 @@
...
+Optionally, an alternative can be followed by a so-called action
+in curly-braces, which specifies the return value of the alternative:
+```
+rule_name[return_type]:
+    | first_alt1 first_alt2 { first_alt1 }
+    | second_alt1 second_alt2 { second_alt1 }
+```
+If the action is omitted, then there are two different possibilities:
+1.  If there's a single name in the alternative, this gets returned.
+2.  If not, a dummy name object gets returned.
```
gvanrossum: Does this also apply to the Python code generator?

docs/grammar.md
```diff
@@ -0,0 +1,90 @@
+Syntax
+------
```
gvanrossum: I wonder if we should include a small example for each documented construct? That usually helps readers grasp the construct quicker than just reading the abstract description.

docs/grammar.md
```diff
@@ -0,0 +1,90 @@
...
+    | second_alt1 second_alt2 { second_alt1 }
+```
+If the action is omitted, then there are two different possibilities:
+1.  If there's a single name in the alternative, this gets returned.
+2.  If not, a dummy name object gets returned.
+
+
+### Variables in the Grammar
+
+An expression can be named.  The name can then be used in the action,
```
gvanrossum: "A subexpression can be named by preceding it with an identifier and an `=` sign."

docs/helpers.md
```diff
@@ -0,0 +1,97 @@
+Helper Structs
+--------------
```
gvanrossum: Explain that these are C types defined in pegen.h.

docs/helpers.md
```diff
@@ -0,0 +1,97 @@
...
+Linked list, optimized for quickly finding a given type.
+
+- type: int, either a token or a rule (rules start at 256)
+- node: NULL or pointer to AST node OR pointer to token object
+- mark: if node != NULL, index into Parser's array of tokens
+- next: NULL or pointer to next Memo structure
+
+##### Token
+
+These are in an array linked from Parser.  (Or linked list???)
```
gvanrossum: It's an array. We use a linked list for `Memo` only. (Maybe we should explain how the data structures fit together? It would be best explained using a diagram.)

docs/helpers.md
```diff
@@ -0,0 +1,97 @@
...
+- type: int, token type (needs only 8 bits)
+- bytes: bytes object
+- lineno, col_offset, end_lineno, end_col_offset: int
+- memo: Pointer to linked list Memo
+
+##### Parser
+
+Also needs to point to a PyArena, used for allocating AST nodes and
+other things.  And maybe some flag indicating there's an allocation
+error and a convention to bail if there is one.
```
gvanrossum: Allocation errors were always dealt with by bailing immediately (Pablo added this).

docs/helpers.md
```diff
@@ -0,0 +1,97 @@
...
+
+- cmpop: cmpop_ty, The comparison operator
+- expr: expr_ty, The expression that gets compared
+
+
+Helper Functions
+----------------
+
+###### `asdl_seq *singleton_seq(Parser *, void *);`
+Creates a single-element asdl_seq* that contains a
```
gvanrossum: That contains a what? :-) I think you forgot to add the argument names to the header, and I recommend putting `asdl_seq*` and `a` in backticks. Also add a period.

docs/helpers.md
lysnikolaou: I don't really have a clear enough picture of how this should look. Should we maybe merge this PR without the diagram and do that later?

docs/helpers.md
lysnikolaou: I'll do that. I also think that the semicolons should be removed.

docs/grammar.md
lysnikolaou: It doesn't. Updated!

docs/grammar.md
lysnikolaou: I was having difficulties coming up with useful examples for lookahead and cut operators.

PR #98, opened 2019-12-08
Refactor rules for function definitions and parameters

lysnikolaou:
This is still a work in progress. `pytest` passes, but I still get a SEGFAULT in `simpy_cpython`, which I can't understand and which I am looking into.

If anyone can have a look and maybe find the source of the SEGFAULT, it would be great! What I know so far is that the file `cpython/Lib/test/test__xxsubinterpreters.py` is the one that cannot be parsed.

Edit: The segfault is out of the way. This still remains a WIP though, as I still have to update the docs for it.


pegen/pegen.c
```diff
@@ -757,3 +757,192 @@ map_targets_to_del_names(Parser *p, asdl_seq *seq)
...
+    }
+
+    return new_seq;
+}
+
+asdl_seq *
+_get_names(Parser *p, asdl_seq *names_with_defaults)
+{
+    int len = asdl_seq_LEN(names_with_defaults);
+    asdl_seq *seq = _Py_asdl_seq_new(len, p->arena);
```
pablogsal: Please, check the return value before populating the element

pegen/pegen.c
```diff
@@ -757,3 +757,192 @@ map_targets_to_del_names(Parser *p, asdl_seq *seq)
...
+        asdl_seq_SET(seq, i, pair->arg);
+    }
+    return seq;
+}
+
+asdl_seq *
+_get_defaults(Parser *p, asdl_seq *names_with_defaults)
+{
+    int len = asdl_seq_LEN(names_with_defaults);
+    asdl_seq *seq = _Py_asdl_seq_new(len, p->arena);
```
pablogsal: Same here

pegen/pegen.c
```diff
@@ -757,3 +757,192 @@ map_targets_to_del_names(Parser *p, asdl_seq *seq)
...
+
+arguments_ty
+make_arguments(Parser *p, asdl_seq *slash_without_default, SlashWithDefault *slash_with_default,
+               asdl_seq *plain_names, asdl_seq *names_with_default, StarEtc *star_etc)
+{
+    asdl_seq *posonlyargs;
+    if (slash_without_default != NULL) {
+        posonlyargs = slash_without_default;
+    } else if (slash_with_default != NULL) {
+        posonlyargs = _join_seqs(p,
```
pablogsal: `_join_seqs` and `_get_names`can fail, check the return value before passing the value arround

pegen/pegen.c
```diff
@@ -757,3 +757,192 @@ map_targets_to_del_names(Parser *p, asdl_seq *seq)
...
+    }
+
+    arg_ty vararg = NULL;
+    if (star_etc != NULL && star_etc->vararg != NULL) {
+        vararg = star_etc->vararg;
+    }
+
+    asdl_seq *kwonlyargs;
+    if (star_etc != NULL && star_etc->kwonlyargs != NULL) {
+        kwonlyargs = _get_names(p, star_etc->kwonlyargs);
```
pablogsal: `_get_names` can fail.

pegen/pegen.c
```diff
@@ -757,3 +757,192 @@ map_targets_to_del_names(Parser *p, asdl_seq *seq)
...
+        posargs = plain_names;
+    } else {
+        posargs = _Py_asdl_seq_new(0, p->arena);
+        if (!posargs)
+            return NULL;
+    }
+
+    asdl_seq *posdefaults;
+    if (slash_with_default != NULL && names_with_default != NULL) {
+        posdefaults = _join_seqs(p,
```
pablogsal: All of these calls can fail (`_join_seqs`, `_get_defaults` ).

pablogsal:
The segfault is because of the result `AST`is malformed and there is an expression node with an invalid value for `_expr_kind`  (52 while the enum only has 26).

lysnikolaou:
@pablogsal Thanks for the review. I will push a new commit ASAP. In the meantime, the segfault is due to the statement `raise ...`, which is not being parse correctly.

lysnikolaou:
Would it be better to open a new PR for the fix or just push to this one? It has to do with the `raise_stmt`, but it is only a single change in the `atom` rule and the addition of a test file with `raise ...`.

pablogsal:
> Would it be better to open a new PR for the fix or just push to this one?

I would say another PR, as this one is already big enough ;)

gvanrossum:
If Pablo thinks this is ready, don't wait for me -- I have a few other things to take care of, and I find it hard to make time for substantial code reviews. Sorry!

gvanrossum:
Yes, I was trying to follow PEP 7 (though maybe not hard enough :-).

lysnikolaou:
@pablogsal Thanks a lot for the review!! This one definitely wasn't an easy one for me! :tada:

Issue #79, opened 2019-11-28
How to differentiate between Load and Store for NAMEs?


lysnikolaou:
I got to implementing the actions for `while` and `for` and found out that, upon parsing a NAME, `name_token` gets called, which always uses `Load`. Except for the quick workaround

```
target: a=NAME { _Py_Name(a->v.Name.id, Store, EXTRA_EXPR(a, a)) }
```
there is unfortunately no other way to use Store for a NAME.

What do you think is the best way to implement this?

gvanrossum:
Create a helper function that takes a Load name object as input and constructs an otherwise equivalent Store name object?

lysnikolaou:
> Create a helper function that takes a Load name object as input and constructs an otherwise equivalent Store name object?

Or changes the `ctx` field in place?

What I haven't understood though is if there is an occurrence of `NAME` in the grammar, which could be either one, according to where the rule got called from.

gvanrossum:
Because of the way the packrat parser cache works, you shouldn't modify anything that's been cached in place, because a backtrack could cause it to be reinterpreted the other way. This is not hypothetical: if you parse a simple expression like `x`, the parser first parsers it as a target, then finds there's no `=` following that, then backtracks and reinterprets it as an expression.

lysnikolaou:
Understood! Hadn't thought about that till now.

PR #100, opened 2019-12-09
Refactor rules for assignments

pablogsal:


data/simpy.gram
```diff
@@ -24,11 +24,96 @@ small_stmt[stmt_ty]:
...
+         EXTRA(a, expr_type, c, expr_type)) }
+    | a=target ':' b=expression {
+         _Py_AnnAssign(SET_CONTEXT(a, Store), b, NULL, 1,
+         EXTRA(a, expr_type, b, expr_type)) }
+    | a=target '=' b=yield_expression {
+         _Py_Assign(singleton_seq(p, SET_CONTEXT(a, Store)), b, NULL,
+         EXTRA(a, expr_type, b, expr_type)) }
+    | (star_targets '=')+ yield_expr
+    | (star_targets '=')+ expressions
+    | a=target '+=' b=yield_expr {
```
pablogsal: We need a way to allow integer values as the types for rules because this is a bit ridiculous.....

pablogsal:
Opened https://github.com/gvanrossum/pegen/issues/101 for the discussion on how to proceed for simple types for rules.

pegen/pegen.c
```diff
@@ -2,6 +2,22 @@
...
 #include "v38tokenizer.h"
 
+int _set_context(expr_ty e, expr_context_ty ctx)
+{
+    // TODO: Check for valid names and the rest
+    // of the possibilities in the e->kind enum.
+    // Also, research a way to propagate errors.
+    switch (e->kind) {
+        case Name_kind:
+            e->v.Name.ctx = ctx;
```
lysnikolaou: I'm not sure, if this applies here, but @gvanrossum suggests [here](https://github.com/gvanrossum/pegen/issues/79#issuecomment-559611541) that modifying something in place could mess up the cache.

pegen/pegen.h
```diff
@@ -82,6 +87,9 @@ CmpopExprPair *cmpop_expr_pair(Parser *, cmpop_ty, expr_ty);
 expr_ty Pegen_Compare(Parser *, expr_ty, asdl_seq *);
 expr_ty store_name(Parser *, expr_ty);
 asdl_seq *map_targets_to_del_names(Parser *, asdl_seq *);
+int _set_context(expr_ty, expr_context_ty);
+AugOperator* augoperator(Parser*, operator_ty type);
+#define SET_CONTEXT(a,b) (_set_context(a, b), a)
```
lysnikolaou: Not sure about this, but I'd move this line up together with the `EXTRA`s.

pegen/pegen.h
```diff
@@ -36,6 +36,10 @@ typedef struct {
     expr_ty expr;
 } CmpopExprPair;
 
+typedef struct {
+    int kind;
```
lysnikolaou: Why `int` and not `operator_ty`?

pablogsal:
Thanks for the review @lysnikolaou. I need to catch up with the new helpers and your comments are very valuable. I have addressed you feedback but the current rules are failing on some multiple assignments:

```
x = y = expr
```

when parsing Cpython code all failing with token NEWLINE:

```
...
./cpython/Tools/scripts/lfcr.py:13:28: error at line 13, col 28, token NEWLINE
./cpython/Tools/scripts/linktree.py:23:48: error at line 23, col 48, token NEWLINE
./cpython/Tools/scripts/lll.py:13:47: error at line 13, col 47, token NEWLINE
./cpython/Tools/scripts/mailerdaemon.py:21:34: error at line 21, col 34, token NEWLINE
./cpython/Tools/scripts/make_ctype.py:18:15: error at line 18, col 15, token NEWLINE
./cpython/Tools/scripts/md5sum.py:8:16: error at line 8, col 16, token NEWLINE
...
```

I need to investigate this, but observations are welcomed :)

pablogsal:
Hummmm.... it seems that this change breaks the parsing:

```
(star_targets '=')+ (yield_expr | expressions) -> "=".star_targets+ '=' (yield_expr | expressions) 
```

I am missing something with the new "." operator?

pablogsal:
Oh, I wonder if what's happening is that `=".star_targets+` is eagerly trying to consume the next '=' and then it fails to consume a new `start_targets` (because the yield_expr | expressions follows) and then it tries something else.

pegen/pegen.c
gvanrossum: Yes I think it applies here. 

PS I am on a phone at a dentist office, code reviews are tough.

lysnikolaou:
> Oh, I wonder if what's happening is that `=".star_targets+` is eagerly trying to consume the next '=' and then it fails to consume a new `start_targets` (because the yield_expr | expressions follows) and then it tries something else.

Oh yeah, that's it. The `gather` rule fails, after it fails to parse the last `=`, which means that the entire alternative fails directly after that.

pablogsal:
> Oh yeah, that's it. The `gather` rule fails, after it fails to parse the last `=`, which means that the entire alternative fails directly after that.

Do we have any helpers to handle `(star_targets '=')+` ?

lysnikolaou:
> Do we have any helpers to handle `(star_targets '=')+` ?

Nope, see my last comment for how I would do it.

pablogsal:
> Nope, see my last comment for how I would do it.

Apologies, I cannot find what comment yo mention. Could you pass a link?

lysnikolaou:
I edited one of my last comments with a suggestion, so you might need a refresh to all of it. Here is it again:

I'd just fall back to something like a=(z=star_targets '=' { z })+ b=(yield_expr | expressions)

pablogsal:
> I'd just fall back to something like a=(z=star_targets '=' { z })+ b=(yield_expr | expressions)

Fantastic idea! I like this approach :)

pegen/pegen.c
pablogsal: Thanks for pointing that out. I have used the new helper `store_name` in f8f48c4

pegen/pegen.h
```diff
@@ -82,6 +87,7 @@ CmpopExprPair *cmpop_expr_pair(Parser *, cmpop_ty, expr_ty);
 expr_ty Pegen_Compare(Parser *, expr_ty, asdl_seq *);
 expr_ty store_name(Parser *, expr_ty);
 asdl_seq *map_targets_to_del_names(Parser *, asdl_seq *);
+AugOperator* augoperator(Parser*, operator_ty type);
```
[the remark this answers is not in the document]
pablogsal: > This is maybe nitpicking, but leaving a space before the star for consistency might be a good idea

Agreed! Thanks for pointing that out. I have corrected it in the last commit.

pablogsal:
> That's of course for another PR, but I think an issue is a good idea.

Opened https://github.com/gvanrossum/pegen/issues/103 to track this

pablogsal:
@gvanrossum Is it okay if we go ahead and merge this with one approval (in this case the one from @lysnikolaou)?

gvanrossum:
Sure!

PR #108, opened 2019-12-11
Script that shows the parse tree for a small program

gvanrossum:
For example,
```
$ scripts/show_parse.py a+b
Module(
    body=[
        Expr(
            value=BinOp(
                left=Name(id="a", ctx=Load()), op=Add(), right=Name(id="b", ctx=Load())
            )
        )
    ],
    type_ignores=[],
)
```


lysnikolaou:
Thanks for doing this! Really helpful!

PR #109, opened 2019-12-11
Refactor various atom rules

### LINKED ISSUE #106: What should the expressions rule return?
Up until now I was under the impression that the `expressions` rule should return either a tuple, if there are multiple expressions or the expression itself, if there only is one.

But trying to refactor the various atom rules, I saw this:
```
set: '{' expressions '}'
```
which means that expressions should return an `asdl_seq *` of all the parsed expressions?

What is the right approach here?

lysnikolaou:
* list
* listcomp
* tuple
* group
* genexp
* set
* setcomp
* dict
* dictcomp

Closes #106.


test/test_ast_generation.py
```diff
@@ -37,6 +37,8 @@ def parser_extension(tmp_path_factory: Any) -> Any:
 
 @pytest.mark.parametrize("filename", PYTHON_SOURCE_FILENAMES)
 def test_ast_generation_on_source_files(parser_extension: Any, filename: PurePath) -> None:
+    if filename == "group.py":
```
pablogsal: Suggestion: We can write an independent test for this additionally, and mark it for now as `xfail` so we don't forget and know when is fixed

test/test_data/set.py
```diff
@@ -0,0 +1 @@
+{1, 2, 3}
```
pablogsal: Suggestion: I would expressions (like `1+2`) in these examples as they can catch a bigger subset of weird behaviour than just number literals.

test/test_data/list_comp.py
```diff
@@ -0,0 +1 @@
+[i for i in a]
```
pablogsal: I think we are missing comprehensions with conditionals in all cases:

```
[i for i in a if something]
```

test/test_data/set_trailing_comma.py
```diff
@@ -0,0 +1 @@
+{1, 2, 3,}
```
pablogsal: Nit: No newline at the end of the file

gvanrossum:
Looks like you have a merge conflict now.

pablogsal:
> Looks like you have a merge conflict now.

Is very easy to solve as we just need to accept all "branches" (this is due to the fact that functions added in this PR are conflicting in position with functions added in old merged PRs).

pegen/pegen.c
```diff
@@ -772,3 +772,53 @@ map_targets_to_del_names(Parser *p, asdl_seq *seq)
     }
     return new_seq;
 }
+
+KeyValuePair *
+key_value_pair(Parser *p, expr_ty key, expr_ty value)
```
pablogsal: Nit: For consistency with other functions in this file, we should add a comment before the function as documentation. Example of what I mean:

```
/* Creates an asdl_seq* where all the elements have been changed to have del as context */
asdl_seq *
map_targets_to_del_names(Parser *p, asdl_seq *seq)	map_targets_to_del_names(Parser *p, asdl_seq *seq)
{
 int len = asdl_seq_LEN(seq);
```

lysnikolaou:
@pablogsal Thanks a lot for the review! Could you maybe have another look?

@gvanrossum Conflicts resolved!

data/simpy.gram
```diff
@@ -340,18 +346,31 @@ atom[expr_ty]:
...
-for_if_clauses: ([ASYNC] 'for' star_targets 'in' expression ('if' expression)*)+
+
+list[expr_ty]:
+    | a='[' b=[star_full_expressions] c=']' { _Py_List(b, Load, EXTRA(a, token_type, c, token_type)) }
+listcomp[expr_ty]:
+    | a='[' b=full_expression c=for_if_clauses d=']' { _Py_ListComp(b, c, EXTRA(a, token_type, d, token_type)) }
+tuple[expr_ty]:
+    | a='(' b=[y=star_full_expression ',' z=[star_full_expressions] { seq_insert_in_front(p, y, z)} ] c=')' {
+        _Py_Tuple(b, Load, EXTRA(a, token_type, c, token_type)) }
+group[expr_ty]: '(' a=(yield_expr | full_expression) ')' { a }
```
pablogsal: As this PR also changes the expressions, would it make sense to add some tests that exercise the `( yield )`  case? We can do them in a different PR

data/simpy.gram
```diff
@@ -211,17 +211,23 @@ class_def: [decorators] 'class' NAME ['(' [arguments] ')'] ':' block
...
+        _Py_Tuple(seq_insert_in_front(p, a, b), Load, EXTRA_EXPR(a, seq_get_tail(NULL, b))) }
+    | a=star_expression ',' { _Py_Tuple(singleton_seq(p, a), Load, EXTRA_EXPR(a, a)) }
+    | star_expression
+star_expression[expr_ty]:
+    | a='*' b=bitwise_or { _Py_Starred(b, Load, EXTRA(a, token_type, b, expr_type)) }
     | expression
 
-star_full_expression: '*' bitwise_or | full_expression
-full_expression: NAME ':=' expression | expression
+star_full_expressions[asdl_seq*]: a=','.star_full_expression+ [','] { a }
```
pablogsal: I think we should add some explicit cases for these rules, because the pattern `"SOMETHING".rulename "SOMETHING"` makes me now suspect (not that this is wrong, but this was failing very subtly on the assignment PR) of the parser eagerly consuming the "SOMETHING".

data/simpy.gram
lysnikolaou: I don't exactly understand, which case you mean? There is the `(yield a)` test. Is there something more than that that needs to be added?

pablogsal: > There is the `(yield a)` test

My bad, I have missed it :S

data/simpy.gram
lysnikolaou: Added a new test file `list_trailing_comma.py` that uses `star_named_expressions` and a trailing comma. I don't think there is a problem. `','.something+` is designed to be the same with `something (',' something)*`, which is what `star_named_expressions` was before. In the assignment case, the rule was actually somewhat different, which is why we missed it at the beginning.

pablogsal: Yup, there is no problem but I still think having a test case for this is slightly better in case we do future refactors or similar. Maybe is just me fearing the wrath of PEG with these patterns :)

pegen/pegen.c
```diff
@@ -857,6 +869,46 @@ name_default_pair(Parser *p, arg_ty arg, expr_ty value)
...
+    }
+    for (int i = 0; i < len; i++) {
+        KeyValuePair *pair = asdl_seq_GET(seq, i);
+        asdl_seq_SET(new_seq, i, pair->value);
+    }
+    return new_seq;
+}
+
+int
+is_async(void *keyword)
```
pablogsal: Nit (feel free to ignore): Should this function be public? If not, can you mark it as `static`? Otherwise, should it make sense to add it to the `doc`?

lysnikolaou: Actually, with #111 we can remove this function altogether.

lysnikolaou:
Now I am getting an error when I run `simpy_cpython`, without changing anything. The error I'm getting is the following:
```
./Include/object.h:541: _Py_NegativeRefcount: Assertion failed: object has negative ref count
<object at 0x55ac8706e920 is freed>
Fatal Python error: _PyObject_AssertFailed
Python runtime state: initialized

Current thread 0x00007f11e41d3280 (most recent call first):
  File "scripts/test_parse_directory.py", line 127 in main
  File "scripts/test_parse_directory.py", line 168 in <module>
Aborted (core dumped)
make: *** [Makefile:51: simpy_cpython] Error 134
```

I updated my local cpython, in order to work on the cpython PR, after quite some time. Could that have be the cause of this?

gvanrossum:
Can you tell me how to reproduce this? Is it possible that the only thing
that changed is the config flags used to build CPython? This error looks
like something that only occurs in debug mode. I presume you used the 3.8
branch?

pablogsal:
> Now I am getting an error when I run `simpy_cpython`, without changing anything. The error I'm getting is the following (the file that causes this is `cpython/Lib/test/ann_module.py`):

I am investigating, I have an idea of what can be the cause of this.

lysnikolaou:
I'm not gonna be online for an hour or so. I'll investigate further and try
to find out what are the exact steps to reproduce this when I get back.

pablogsal:
Found the problem:

https://github.com/gvanrossum/pegen/pull/115

Basically, it was a reference count issue in the implementation of `store_name` introduced in commit 83719e4d

lysnikolaou:
> I presume you used the 3.8 branch?

Do you mean running `simpy_cpython` against the 3.8 branch? It should also succeed on master, or not?

pablogsal:
It seems that the problem is caused by:

```
(pars): bool = True
```

if you remove the parenthesis the segfault disappears:

```
~/github/pegen master* ⇡
venv ❯ python -c 'import parse; parse.parse_string("(pars): bool = True")'
./Include/object.h:541: _Py_NegativeRefcount: Assertion failed: object has negative ref count
<object at 0x5648ba99a470 is freed>
Fatal Python error: _PyObject_AssertFailed
Python runtime state: initialized

Current thread 0x00007ff029139740 (most recent call first):
  File "<string>", line 1 in <module>
[1]    21198 abort (core dumped)  python -c 'import parse; parse.parse_string("(pars): bool = True")'

~/github/pegen master* ⇡
venv ❯ python -c 'import parse; parse.parse_string("pars: bool = True")' 

```

pablogsal:
Ok, I think I understand now. What's happening is that the `store_name()` call is receiving an incorrect type. When there is parenthesis, the target returns a _Py_Tuple instead of a NAME:

```
 t_atom[expr_ty]:
      | NAME
      | a='(' b=[targets] c=')' { _Py_Tuple(b, Load, EXTRA(a, token_type, c, token_type)) }
      | a='[' b=[targets] c=']' { _Py_List(b, Load, EXTRA(a, token_type, c, token_type)) }
```

so when we do

```
  assignment:
      | a=target ':' b=expression '=' c=yield_expression {
           _Py_AnnAssign(store_name(p, a), b, c, 1, EXTRA_EXPR(a, c)) }
```

That `store_name` thinks is receiving a name, but indeed is receiving a `_Py_Tuple`
```

lysnikolaou:
I was just wondering if that's maybe incorrect. It seems, that this rule is incorrect. Basically, if there's a parenthesis, then there are two different possible return values. A tuple if the parentheses contain more than one element or the element itself, if not.

lysnikolaou:
It seems that the `simpy_cpython` is okay on Travis. Should I merge this PR and fix targets in a new one?

pablogsal:
> I was just wondering if that's maybe incorrect. It seems, that this rule is incorrect. Basically, if there's a parenthesis, then there are two different possible return values. A tuple if the parentheses contain more than one element or the element itself, if not.

I have created https://github.com/gvanrossum/pegen/pull/116. Could you take a look?

pablogsal:
> It seems that the `simpy_cpython` is okay on Travis. Should I merge this PR and fix targets in a new one?

I think that is ok to merge this as the error is unrelated.

PR #110, opened 2019-12-11
Add capability of parsing using arbitrary grammar to show_parse script

gvanrossum:
Use the same -g flag as test_parse_directory.py (but with less error checking for now)

### DIFF
--- a/scripts/show_parse.py
+++ b/scripts/show_parse.py
@@ -30,6 +30,7 @@ import sys
 import tempfile
 
 parser = argparse.ArgumentParser()
+parser.add_argument("-g", "--grammar-file", help="grammar to use (default: use the ast module)")
 parser.add_argument("-v", "--verbose", action="store_true", help="show line/column numbers")
 parser.add_argument("program", nargs="+", help="program to parse (will be concatenated)")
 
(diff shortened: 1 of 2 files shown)


scripts/show_parse.py
```diff
@@ -55,7 +56,20 @@ def print_parse(source: str, verbose: bool = False) -> None:
 
 def main() -> None:
     args = parser.parse_args()
-    print_parse(" ".join(args.program), args.verbose)
+    program = " ".join(args.program)
+    if args.grammar_file:
+        sys.path.insert(0, os.curdir)
+        from pegen.build import build_parser_and_generator
+
+        build_parser_and_generator(args.grammar_file, "pegen/parse.c")
```
lysnikolaou: Due to passing "pegen/parse.c" as the filename, one can only execute the script from the root directory of the project. If not, then just a `FileNotFoundError` is thrown, which might be a bit misleading. Then again we may be the only ones using the script, so it's okay.

gvanrossum: The same is done by two other scripts; I'd rather not complicate the script without reason. If we find this is confusing users we can do something better (like computing the correct directory based on sys.argv[0]).

PR #111, opened 2019-12-11
Change metagrammar to allow for ternary operator

lysnikolaou:


lysnikolaou:
Don't know exactly how or where this is handled, but it seems that spaces before the "?" are allowed.

PR #112, opened 2019-12-11
Add --diff flag to show (only) the diff between our parse and ast.parse

gvanrossum:


scripts/show_parse.py
```diff
@@ -56,6 +70,8 @@ def print_parse(source: str, verbose: bool = False) -> None:
 
 def main() -> None:
     args = parser.parse_args()
+    if args.diff and not args.grammar_file:
+        parser.error("--diff requires --grammar-file")
```
pablogsal:
```suggestion
        parser.error("--diff/-d requires --grammar-file/-g")
```

gvanrossum: Good idea, did this as a separate commit (no PR) since I already merged. :-)

Issue #113, opened 2019-12-11
Populate text attribute of SyntaxError


gvanrossum:
In order for SyntaxError to be formatted with a nice display of the line, like this,
```
  File "<unknown>", line 1
    1/
     ^
SyntaxError: invalid syntax
```
we must set the `text` attribute to the line containing the error.

PR #114, opened 2019-12-11
Create a function to raise syntax errors

pablogsal:
Create a function to be reused to create and populate a syntax error
from the available information in the current parser state.

closes: #113

### DIFF
--- a/pegen/pegen.c
+++ b/pegen/pegen.c
@@ -20,7 +20,7 @@ raise_syntax_error(Parser *p, const char *errmsg, ...)
     if (!errstr) {
         goto error;
     }
-    if (filename) {
+    if (p->tok->filename) {
         filename = p->tok->filename;
         loc = PyErr_ProgramTextObject(filename, t->lineno);
         if (!loc) {
--- a/pegen/pegen.c
+++ b/pegen/pegen.c
@@ -41,7 +41,7 @@ raise_syntax_error(Parser *p, const char *errmsg, ...)
     } else {
         Py_INCREF(Py_None);
         filename = Py_None;
-        loc = PyUnicode_DecodeUTF8(p->tok->buf, (p->tok->cur)-(p->tok->buf), "replace");
+        loc = PyUnicode_FromString(p->tok->buf);
         if (!loc) {
             goto error;
         }
--- a/pegen/pegen.c
+++ b/pegen/pegen.c
@@ -2,6 +2,62 @@
 #include "pegen.h"
 #include "v38tokenizer.h"
 
+int
+raise_syntax_error(Parser *p, const char *errmsg, ...)
+{
+    // TODO: comvert from bytes offset to character offset
+    PyObject *value = NULL;
+    PyObject *errstr = NULL;
+    PyObject *loc = NULL;
+    PyObject *tmp = NULL;
+    PyObject* filename = NULL;
+    Token *t = p->tokens[p->fill - 1];
+    va_list va;
+
+    va_start(va, errmsg);
+    errstr = PyUnicode_FromFormatV(errmsg, va);
+    va_end(va);
+    if (!errstr) {
+        goto error;
+    }
+    if (filename) {
+        filename = p->tok->filename;
+        loc = PyErr_ProgramTextObject(filename, t->lineno);
+        if (!loc) {
+            Py_INCREF(Py_None);
+            loc = Py_None;
+        }
+    } else {
+        Py_INCREF(Py_None);
+        filename = Py_None;
+        loc = PyUnicode_DecodeUTF8(p->tok->buf, (p->tok->cur)-(p->tok->buf), "replace");
+        if (!loc) {
+            goto error;
+        }
+    }
+    tmp = Py_BuildValue("(OiiN)", filename, t->lineno, t->col_offset + 1, loc);
+    if (!tmp) {
+        goto error;
+    }
+    value = PyTuple_Pack(2, errstr, tmp);
+    Py_DECREF(tmp);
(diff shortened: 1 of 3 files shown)


pablogsal:
Before this PR:

```
python -c 'import parse; parse.parse_string("1-+*2")'               
  File "<string>", line 1
SyntaxError: error at line 1, col 4, token STAR

```

After this PR:

```
python -c 'import parse; parse.parse_string("1-+*2")'               
  File "<string>", line 1
    1-+*
       ^
SyntaxError: invalid syntax
```

pegen/pegen.c
```diff
@@ -2,6 +2,62 @@
...
+    Token *t = p->tokens[p->fill - 1];
+    va_list va;
+
+    va_start(va, errmsg);
+    errstr = PyUnicode_FromFormatV(errmsg, va);
+    va_end(va);
+    if (!errstr) {
+        goto error;
+    }
+    if (filename) {
```
gvanrossum: How can this condition be true? It's initialized to NULL and not set between there and here. I'm guessing you meant `if (p->tok->filename)` ? (And could that be `Py_None`? Could `p->tok` be NULL?)

pegen/pegen.c
```diff
@@ -2,6 +2,62 @@
 #include "pegen.h"
 #include "v38tokenizer.h"
 
+int
+raise_syntax_error(Parser *p, const char *errmsg, ...)
+{
+    // TODO: comvert from bytes offset to character offset
```
gvanrossum: Is it very onerous to just do this, rather than leaving a TODO?

pegen/pegen.c
pablogsal: Yeah, that was meant to be `(p->tok->filename)`.

> Could p->tok be NULL?

The previous code was using `p->tok` directly is a similar fashion so I will assume at this point it won't be true, but we can add some early exit or asserts. What do you prefer?

gvanrossum: No need to add checks in that case.

pegen/pegen.c
pablogsal: No, but I was not sure I understood what field was the comment referring to and I wanted to discuss this. What changes does this `TODO` require?

pablogsal:
> Can you come up with a way to test this? (Preferably both the "from file" and "from string" cases.)

Yup, will add some test cases.

pegen/pegen.c
gvanrossum: I assume it's about column numbers being measured in bytes vs. characters, if the file has an encoding? You may have to attempt to repro this by creating a file or string with a unicode identifier followed by a syntax error to see what offset is displayed.

pablogsal:
> Try these tests with Unicode names?

I have added a new individual test for this case

pegen/pegen.c
```diff
@@ -30,12 +41,13 @@ raise_syntax_error(Parser *p, const char *errmsg, ...)
     } else {
         Py_INCREF(Py_None);
         filename = Py_None;
-        loc = PyUnicode_DecodeUTF8(p->tok->buf, (p->tok->cur)-(p->tok->buf), "replace");
+        loc = PyUnicode_FromString(p->tok->buf);
```
gvanrossum: What evidence do we have that `tok->buf` is always UTF8? It makes sense for it to be, but I'm not exactly thrilled about how the code in tokenizer.c is written, it's impossible to follow... Unless you have a quick insight this is not actionable, no need to research it further.

pegen/pegen.c
```diff
@@ -30,12 +41,13 @@ raise_syntax_error(Parser *p, const char *errmsg, ...)
...
         if (!loc) {
             goto error;
         }
     }
-    tmp = Py_BuildValue("(OiiN)", filename, t->lineno, t->col_offset + 1, loc);
+    Py_ssize_t col_offset = byte_offset_to_character_offset(loc, t->col_offset + 1);
```
gvanrossum: This should add 1 to the result of the call, not to the input -- if there's a multibyte character at the offset, the +1 will cause PyUnicode_DecodeUTF8 to fail. And I'd call the local variable col_number, since it's 1-based.

test/test_c_parser.py
```diff
@@ -267,3 +268,38 @@ def test_ternary_operator(tmp_path: PurePath) -> None:
...
+    start: expr+ NEWLINE? ENDMARKER
+    expr: NAME
+    """
+    grammar = parse_string(grammar_source, GrammarParser)
+    extension = generate_parser_c_extension(grammar, tmp_path)
+    try:
+        extension.parse_string(text)
+    except SyntaxError as e:
+        tb = traceback.format_exc()
+    assert 'File "<string>", line 1' in tb
```
gvanrossum: Shouldn't those be double quotes? Try "make black".

pegen/pegen.c
pablogsal: > What evidence do we have that `tok->buf` is always UTF8?

I based this implementation in the text of PEP 263:

> Python's tokenizer/compiler combo will need to be updated to work as follows:
>* read the file
>* decode it into Unicode assuming a fixed per-file encoding
>* convert it into a UTF-8 byte string
>* tokenize the UTF-8 content
>* compile it, creating Unicode objects from the given Unicode data and creating string objects from the Unicode literal data by first reencoding the UTF-8 data into 8-bit string data using the given file encoding

It seems that at the point the tokenizer acts, is always utf8

gvanrossum:
Great!

pablogsal:
Thanks for the thorough review, @gvanrossum!

PR #115, opened 2019-12-12
Fix reference count in store_name()

pablogsal:
In the store_name() function we are creating a new _Py_Name struct that
takes the id field from the old one. The id field is a Python object and
therefore to take propper ownership, the reference count needs to be
increased.

### DIFF
--- a/pegen/pegen.c
+++ b/pegen/pegen.c
@@ -804,6 +804,7 @@ store_name(Parser *p, expr_ty load_name)
     if (!load_name) {
         return NULL;
     }
+    Py_INCREF(load_name->v.Name.id);
     return _Py_Name(load_name->v.Name.id,
                     Store,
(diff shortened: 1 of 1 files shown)


pablogsal:
Let me run some extra tests to make sure I am not missing something and I will merge this

lysnikolaou:
Definitely my bad there! Sorry!

pablogsal:
Hummmm, there is something missing still. I know that this object is the problem (as proof, using a fresh string in the new struct instead solves the problem) but the `Py_Incref` does not fully work because of something with the arena.

pablogsal:
> Definitely my bad there! Sorry!

Don't worry, this is known to be very tricky

pablogsal:
I'm closing this until I get the full picture because I am clearly missing something.

PR #116, opened 2019-12-13
Handle complex annotations in assingment rule

pablogsal:


pegen/pegen.c
```diff
@@ -799,14 +799,33 @@ Pegen_Compare(Parser *p, expr_ty expr, asdl_seq *pairs)
...
+    if (!node) {
         return NULL;
     }
-    return _Py_Name(load_name->v.Name.id,
-                    Store,
-                    EXTRA_EXPR(load_name, load_name));
+
+    switch(node->kind) {
+        case Name_kind:
+            return _Py_Name(node->v.Name.id,
```
lysnikolaou: Do we still need to Py_INCREF here?

pablogsal: Unclear, I still need to understand how this "borrow" works but I suspect it *may* be ok because the arena is holding it alive and that lives as long as the ast nodes live.

test/test_data/annotated_assignment_with_parens.py
```diff
@@ -0,0 +1 @@
+(paren): int = 1 + 2
```
lysnikolaou: This will probably fail as well. The correct AST for this only contains the name itself, but we generate the tuple.

pablogsal:
Oh, it seems that python expects the name inside the tuple and not the tuple itself :(

lysnikolaou:
This is a bit tricky to get right! If there's a single target in the parentheses then it should be returned as is, not in a tuple. At the same time, if there's a comma after that single target, a tuple should be returned.

pablogsal:
> At the same time, if there's a comma after that single target, a tuple should be returned.

You mean `(x,): int = 3,4`? That is not valid syntax I think

lysnikolaou:
Yes, but `target` is used for simple assignments without annotations as well and `(x,) = y` is valid syntax.

lysnikolaou:
> Yes, but `target` is used for simple assignments without annotations as well and `(x,) = y` is valid syntax.

Oops, that's wrong star_targets is used for assignments without annotations.

pablogsal:
@lysnikolaou Check this version of the PR, I think that this is the simplest solution to this problem. Sorry for the confusion, I clearly need some rest today :P

pablogsal:
>  but I'm a bit hesitant about extracting the first element of the tuple within store_name

Agreed, but if we receive a tuple, is there a case when there can be more elements? Or do you mean that this should be done in a different function?

> In any case, the `targets` rule is going to be the next thing I work on, so it's okay if it sits there for a while like this.

I suspect that refactoring the `targets` will involve a much more complicated process. For example, consider this:

```
x[2][1][4][6]: int
```

that generates this ast in Cpython:

```
Module(
    body=[
        AnnAssign(
            target=Subscript(
                value=Subscript(
                    value=Subscript(
                        value=Subscript(
                            value=Name(id="x", ctx=Load()),
                            slice=Index(value=Constant(value=2, kind=None)),
                            ctx=Load(),
                        ),
                        slice=Index(value=Constant(value=1, kind=None)),
                        ctx=Load(),
                    ),
                    slice=Index(value=Constant(value=4, kind=None)),
                    ctx=Load(),
                ),
                slice=Index(value=Constant(value=6, kind=None)),
                ctx=Store(),
            ),
            annotation=Name(id="int", ctx=Load()),
            value=None,
            simple=0,
        )
    ],
    type_ignores=[],
)
```

this means that we would need to copy the entire structure and change the context everywhere if I am correct.

lysnikolaou:
> Agreed, but if we receive a tuple, is there a case when there can be more elements? Or do you mean that this should be done in a different function?

Yes, when parsing `del (a, b)` the `targets` rule, should return a tuple.

> I think the long-term solution is implementing correctly set_context.

That's true, maybe something like `map_names_to_del_names` is what we need to implement in the long term, but for store and for more than names, lists and tuples.

pablogsal:
@lysnikolaou In 164869b I have refactored the function into something different to keep it isolated, tell me if you prefer it this way for now.

lysnikolaou:
> I think this may be unique in the long run as well because it needs to transform tuples into names and assignment does not allow tuples of more than 1 element.

I think so, too. I guess I made things even more confusing by mentioning `del` in my previous comment and even confused myself in thinking that the `targets` rule will not a return a tuple anymore, which it will. Sorry for that!

In any case, I like this PR better in its current state, with a special function for assignments. If we decide to implement something more general and also begin with the refactorings for `targets`, we can reconsider this.

pablogsal:
> In any case, I like this PR better in its current state, with a special function for assignments. If we decide to implement something more general and also begin with the refactorings for `targets`, we can reconsider this.

For now I have added a way to raise exceptions (we still need a better mechanism in general to signal problems in helpers). This is how it behaves with invalid tuples:

```
~/github/pegen fix*
venv ❯ python -c 'import parse; parse.parse_string("(x,y): int")'
SyntaxError: Only single target (not tuple) can be annotated

~/github/pegen fix*
venv ❯ python -c 'import parse; parse.parse_string("(): int")' 
SyntaxError: Only single target (not tuple) can be annotated
```

pegen/pegen.c
```diff
@@ -2,6 +2,19 @@
 #include "pegen.h"
 #include "v38tokenizer.h"
 
+static PyObject *
+create_dummy_identifier(Parser *p) {
```
lysnikolaou: Nit: Include leading underscore?

pablogsal: Done!

pablogsal:
I am going to proceed and merge this. We probably need to discuss again some of the things we have discussed here when we talk about `target` :)

Thank you very much for the patience and the thorough review and apologies for the confusion. Is 1:33 AM in London and I should call this a day :)

lysnikolaou:
>  I am going to proceed and merge this. We probably need to discuss again some of the things we have discussed here when we talk about target :)

True!

> Thank you very much for the patience and the thorough review and apologies for the confusion. Is 1:33 AM in London and I should call this a day :)

Thanks a lot for doing this! It is 2.33 AM in Berlin, so I think I need to get some rest as well.

Issue #107, opened 2019-12-10
Group line and column metadata needs special handling


lysnikolaou:
When writing the action for the `group` rule, I found out that parsing something like `(yield a)` returns the following AST object:
```
Module(
    body=[
        Expr(
            value=Yield(
                value=Name(
                    id="a", ctx=Load(), lineno=1, col_offset=7, end_lineno=1, end_col_offset=8
                ),
                lineno=1,
                col_offset=1,
                end_lineno=1,
                end_col_offset=8,
            ),
            lineno=1,
            col_offset=0,
            end_lineno=1,
            end_col_offset=9,
        )
    ],
    type_ignores=[],
)
```

That means that the `Expr` node's start column offset is 0, although the `yield` node's is 1. The correct action for `group` is thus:
```
group[expr_ty]: '(' a=(yield_expr | full_expression) ')' { a }
```
but the surrounding `Expr` node needs to have a different column offset and I can't seem to come up with a way of handling this, so any ideas are welcome!

gvanrossum:
Hm, so `Expr` is the node type for an expression statement. I suspect the problem is not specific to `group` but occurs for `dict` and `set` as well, and perhaps others. But can it happen for interior nodes?

I guess this is another reason why I think we may have to come up with a way to gather line/column information out of band, rather than passing it to the node constructors using `EXTRA()`.

(PS. I reformatted your code so I could read it. The easiest way to do it that I've found is to write it to a file and run black over it... I'm weaponizing this as we speak. :-)

lysnikolaou:
> I suspect the problem is not specific to `group` but occurs for `dict` and `set` as well, and perhaps others.

Having refactored all of the `atom` rules(list, tuple, dict, comprehensions, set and so on) it is actually specific only to `group`, because it's the only one where the metadata of the inner node is different to that of the outer statement. And that makes sense, in a `dict` for example, the `dict` node starts at the opening brace and ends at the closing brace. The problem with group is that the parentheses are not part of the expression, but are only used for, well, grouping.

PS Very good idea! I was trying to do it using `pprint` but couldn't get it formatted the way I wanted.

gvanrossum:
Ah, so it's because the group disappears. Can we introduce a helper object? Or does it have to be an existing `expr_ty`?

I propose not to worry about this -- I'm not sure it's worth reproducing this particular behavior. OTOH maybe the solution really is to take the line/col info out of band.

PS. https://github.com/gvanrossum/pegen/pull/108

[...]

PR #118, opened 2019-12-13
Move the definition of the inline functions to pegen.c

pablogsal:
As pegen.h is included both in pegen.c and the generated parser file,
having the definition of the inline functions in the header file can
lead to double definition errors if the extension is compiled without
optimizations (-O0) due to the fact that the function is not inlined in
such scenario and then it becomes a regular function that suffers from
double definition.

### DIFF
--- a/pegen/pegen.h
+++ b/pegen/pegen.h
@@ -121,23 +121,23 @@ arguments_ty empty_arguments(Parser *);
 AugOperator *augoperator(Parser*, operator_ty type);
 expr_ty construct_assign_target(Parser *p, expr_ty node);
 
-inline int expr_type_headline(expr_ty a) { return a->lineno; }
-inline int expr_type_headcol(expr_ty a) { return a->col_offset; }
-inline int expr_type_tailline(expr_ty a) { return a->end_lineno; }
-inline int expr_type_tailcol(expr_ty a) { return a->end_col_offset; }
-inline int stmt_type_headline(stmt_ty a) { return a->lineno; }
-inline int stmt_type_headcol(stmt_ty a) { return a->col_offset; }
-inline int stmt_type_tailline(stmt_ty a) { return a->end_lineno; }
-inline int stmt_type_tailcol(stmt_ty a) { return a->end_col_offset; }
-inline int excepthandler_type_headline(excepthandler_ty a) { return a->lineno; }
-inline int excepthandler_type_headcol(excepthandler_ty a) { return a->col_offset; }
-inline int excepthandler_type_tailline(excepthandler_ty a) { return a->end_lineno; }
-inline int excepthandler_type_tailcol(excepthandler_ty a) { return a->end_col_offset; }
-inline int token_type_headline(Token *a) { return a->lineno; }
-inline int token_type_headcol(Token *a) { return a->col_offset; }
-inline int token_type_tailline(Token *a) { return a->end_lineno; }
-inline int token_type_tailcol(Token *a) { return a->end_col_offset; }
-inline int alias_type_headline(PegenAlias *a) { return a->lineno; }
-inline int alias_type_headcol(PegenAlias *a) { return a->col_offset; }
-inline int alias_type_tailline(PegenAlias *a) { return a->end_lineno; }
-inline int alias_type_tailcol(PegenAlias *a) { return a->end_col_offset; }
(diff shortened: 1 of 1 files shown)


pablogsal:
If you want to see the problem by yourself, this is a simple reproducer:

```
(pegen) ❯ CFLAGS='-O0 -g' python scripts/show_parse.py -g data/simpy.gram 'del x[1][2][3]' --diff
token_name(int type)
^
1 warning generated.
  File "scripts/show_parse.py", line 114, in <module>
    main()
  File "scripts/show_parse.py", line 91, in main
    from pegen.parse import parse_string  # type: ignore[import]
ImportError: dlopen(./pegen/parse.cpython-38-darwin.so, 2): Symbol not found: _alias_type_tailcol
  Referenced from: ./pegen/parse.cpython-38-darwin.so
  Expected in: flat namespace
 in ./pegen/parse.cpython-38-darwin.so
```

pablogsal:
> But does the compiler still inline the functions when called from the generated file? How can it even do that?

Because those functions are marked as 'inline' in the header and the header is included in the generated C file before the rest of the code, so when the compiler (with optimizations) see the calls to those functions, will inline the code instead of generating a symbol and a call. The same thing applies when creating the compilation unit for `pegen.c`. Those functions will never generate symbols unless the optimization level forbids inlining.

lysnikolaou:
> Because those functions are marked as 'inline' in the header

But they're not. Or am I missing something?

pablogsal:
> But they're not. Or am I missing something?

They are:

https://github.com/gvanrossum/pegen/blob/master/pegen/pegen.h#L124

pablogsal:
Oh, sorry, I think I misunderstood the question. The question is "how the compiler inline the function AFTER the PR is done?

lysnikolaou:
I thought that the question refers to the state of things after this PR, where you've removed the `inline` keyword from the function declarations in the header.

pablogsal:
In that case, the answer is that it does not inline them, it just ignores the keyword (is still considered a hint):

```C
nm pegen/parse.cpython-38-darwin.so | grep expr_type_headline
0000000000001050 T _expr_type_headline
```

pablogsal:
I can remove the `inline` from the `pegen.c` to avoid confusion if you wish. Or I can transform them into macros and leave them in the header.

pablogsal:
Oh, another thing we can do is mark them as `static inline`. This way every compilation unit will get its own copy and still inline them

lysnikolaou:
Converting them to macros wouldn't be the best idea here. We made them into inline functions, in order for the C type checker to report an error in case we mess up the call to `EXTRA`.

pablogsal:
Oh, another thing we can do is mark them as `static inline`. This way every compilation unit will get its own copy and still inline them

lysnikolaou:
Is there an actual use case, where you need to turn off all optimizations? 

The thing is, that if Guido's suggestion on #107 proves to be sucessful, which it in most likelihood will, then we will get rid of these function as soon as we move to 3.8.1, right? Since they won't be around that long, couldn't we just ignore this and rely on not turning off optimizations?

pablogsal:
> Is there an actual use case, where you need to turn off all optimizations?

The debug build of CPython usually is compiled with -O0 for extra debugging and extension s inherit the CFLAGS

pablogsal:
> Since they won't be around that long, couldn't we just ignore this and rely on not turning off optimizations?

Yep, I would be ok with that in that case.

pablogsal:
Closing this as discussed now the solution of #107 landed in 3.8.

gvanrossum:
I'm not sure how the solution of #107 helps with your inline problem?

pablogsal:
> I'm not sure how the solution of #107 helps with your inline problem?

According to @lysnikolaou:

> if Guido's suggestion on #107 proves to be sucessful, which it in most likelihood will, then we will get rid of these function as soon as we move to 3.8.1

So if those functions will go away soon, it does not matter that much. If that is not the case, we can just declare them as `static inline` in the header so every compilation unit gets its own copy when there are no optimizations.

gvanrossum:
Oh I see, that was my [suggestion](https://github.com/gvanrossum/pegen/issues/107#issuecomment-564647753) to collect line/column info out of band.

PR #120, opened 2019-12-13
Refactor rules for lambda definitions

pablogsal:


pablogsal:
Notice that we have a problem that I think needs to be solved in a different PR when refactoring the `target` rule. The problem is that this:

```
lambda: x
```

is interpreted by the parser as an annotation:

```
(pegen) ❯ python scripts/show_parse.py -g data/simpy.gram 'lambda : 42'
token_name(int type)
^
1 warning generated.
# Parsed using data/simpy.gram
Module(
    body=[
        AnnAssign(
            target=Name(id="lambda", ctx=Store()),
            annotation=Constant(value=42, kind=None),
            value=None,
            simple=1,
        )
    ],
    type_ignores=[],
)
```

pablogsal:
One (ugly) way of solving this may be this:

```diff
diff --git a/data/simpy.gram b/data/simpy.gram
index c17c88e..ac0f95f 100644
--- a/data/simpy.gram
+++ b/data/simpy.gram
@@ -23,12 +23,13 @@ small_stmt[stmt_ty]:
     | e=expressions { _Py_Expr(e, EXTRA_EXPR(e, e)) }
 compound_stmt[stmt_ty]: if_stmt | while_stmt | for_stmt | with_stmt | try_stmt | function_def | class_def

+reserved_target: 'lambda'
 # NOTE: yield_expression may start with 'yield'; yield_expr must start with 'yield'
 assignment:
-    | a=target ':' b=expression '=' c=yield_expression {
+    | !reserved_target a=target ':' b=expression '=' c=yield_expression {
          _Py_AnnAssign(construct_assign_target(p, a), b, c,
          a->kind == Name_kind ? 1 : 0, EXTRA_EXPR(a, c)) }
-    | a=target ':' b=expression {
+    | !reserved_target a=target ':' b=expression {
          _Py_AnnAssign(construct_assign_target(p, a), b, NULL,
          a->kind == Name_kind ? 1 : 0, EXTRA_EXPR(a, b)) }
     | a=(z=star_targets '=' { z })+ b=(yield_expr | expressions) {
```

lysnikolaou:
Thanks for doing this! I had already started doing the necessary
refactorings for this, so I'm wondering if we should have some way of
marking something as work in progress, in order to know who is working on
what? Maybe start an issue with what we're currently working on? Although I
don't remember where, @gvanrossum proposed somewhere to create a
spreadsheet for this, which is also a good idea.

>

gvanrossum:
Well the tokenizer adapter ought to do reserved words, like the old parser.

pablogsal:
Given that this collision is sort of unique for the type annotations, would you be ok to go ahead with the `!'lambda' a=target ':' b=expression` version for the annotations until we have a more robust way of handling the general case?

pablogsal:
> I had already started doing the necessary refactorings for this, so I'm wondering if we should have some way of marking something as work in progress, in order to know who is working on what? Maybe start an issue with what we're currently working on? Although I don't remember where, @gvanrossum proposed somewhere to create a spreadsheet for this, which is also a good idea.

Apologies por the inconvenience! Yes, I think this will be a great idea because now the chance of doing repeated work increases as less and less rules are left. What do you prefer? We can make an issue per rule and assign it to us when we are working on it and we could use it to discuss details there. The spreadsheet can also work for me :)

gvanrossum:
Yeah, the `!'lambda'` hack is fine (I've used similar hacks in a few other places) but long term I think we should reinstate a list of words that are *always* reserved. I just am not sure I want to derive that list from the grammar, like pgen does. (Mostly because I wish we could just write `'async' 'def'` and it would do the right thing, and I have some plans for other non-reserved words in the future.)

lysnikolaou:
> Apologies for the inconvenience!

No need to apologize! :smile: 

> We can make an issue per rule and assign it to us when we are working on it and we could use it to discuss details there. The spreadsheet can also work for me :)

I think issues would work better for me, since I find it easier to have everything in one place (Github). I'll create them some time later today if you and @gvanrossum don't have any objections.

gvanrossum:
Yeah, let's use issues to keep track of this.
-- 
--Guido (mobile)

PR #121, opened 2019-12-14
Refactor EXTRA macro and extract the line and col info out of band

lysnikolaou:
Here's what I've done so far while working on extracting lineno/col_offset out of band. The whole test suite passes with a build of the current 3.8 branch.

I had to change some things in parameters and function_defs, in order to solve the problems listed [here](https://github.com/gvanrossum/pegen/issues/107#issuecomment-565733474). On that same post, I also explain that there is a really serious runtime increase after these changes to anywhere in the range 188-244 seconds. This could be due to the introduced `while` loop that runs for every parsed construct, but I don't really see why that makes such a big difference. Any ideas on how to fight this? 

Update: The runtime increase is not reproducible on Travis.


gvanrossum:
I cannot repro this. After checking out the branch of #121, I still get
```
`/usr/bin/which python3.8` scripts/test_parse_directory.py \
		-g data/simpy.gram \
		-d "./cpython" \
		--short \
		--exclude "*/test2to3/*" \
		--exclude "*/test2to3/**/*" \
		--exclude "*/bad*" \
		--exclude "*/lib2to3/tests/data/*"
token_name(int type)
^
1 warning generated.
Checked 1,641 files, 749,570 lines, 27,622,497 bytes in 5.885 seconds.
That's 127,366 lines/sec, or 4,693,586 bytes/sec.
Memory stats:
  rss         :        110 MiB
  vms         :       4251 MiB
  maxrss      :        110 MiB

real	0m8.980s
user	0m7.891s
sys	0m0.434s
```
which is not significantly slower than on master.

But maybe the Python binary you're using to run this was configured with `--with-pydebug`? That might slow things down?

gvanrossum:
(Or `--with-trace-refs` even more.)

gvanrossum:
(Also the pytest failures are due to the column offset for `elif` -- what shall we do about that?

gvanrossum:
Hm, with a debug Python it's 15 seconds, i.e. about a 3x slowdown, which is reasonable; I still can't repro your ~200 seconds.

lysnikolaou:
Like a good computer scientist does 😄, I restarted my PC and the runtime increase was fixed. Turns out it had something to do with my local environment.

For what it's worth, the binary I'm using was configured `--with-pydebug --without-pymalloc --with-valgrind` and is 3.8.1rc1+. I think I built after the most recent commit to the 3.8 branch. Running `simpy_cpython` against the current CPython master branch with pegen's master and with #121's branch takes about 9 seconds

lysnikolaou:
> Also the pytest failures are due to the column offset for `elif` -- what shall we do about that?

I guess we'll just have to wait for 3.8.1 to be released and integrated into Travis. The purpose of the PR was for you all to have a picture of how it looks and maybe comment on whether I have missed something or whether there is something that could be done better.

data/simpy.gram
```diff
@@ -52,153 +52,153 @@ augassign[AugOperator*]:
     | '//=' {augoperator(p, FloorDiv)}
 
 global_stmt[stmt_ty]: a='global' b=','.NAME+ {
-    _Py_Global(map_names_to_ids(p, b), EXTRA(a, token_type, seq_get_tail(NULL, b), expr_type)) }
+    _Py_Global(map_names_to_ids(p, b), EXTRA) }
```
gvanrossum: Here (for example) we no longer need to name the `'global'` token.

data/simpy.gram
```diff
@@ -52,153 +52,153 @@ augassign[AugOperator*]:
...
+    _Py_Nonlocal(map_names_to_ids(p, b), EXTRA) }
 
-yield_stmt[stmt_ty]: y=yield_expr { _Py_Expr(y, EXTRA_EXPR(y, y)) }
+yield_stmt[stmt_ty]: y=yield_expr { _Py_Expr(y, EXTRA) }
 
 assert_stmt[stmt_ty]:
-    | a='assert' b=expression ',' c=expression { _Py_Assert(b, c, EXTRA(a, token_type, c, expr_type)) }
-    | a='assert' b=expression { _Py_Assert(b, NULL, EXTRA(a, token_type, b, expr_type)) }
+    | a='assert' b=expression ',' c=expression { _Py_Assert(b, c, EXTRA) }
+    | a='assert' b=expression { _Py_Assert(b, NULL, EXTRA) }
```
gvanrossum: Can this be collapsed?
```
    'assert' a=expression b=[',' c=expression { c }] { _Py_Assert(a, b, EXTRA) }
```

data/simpy.gram
```diff
@@ -52,153 +52,153 @@ augassign[AugOperator*]:
...
+
+function_def_raw[stmt_ty]:
+    | s=ASYNC 'def' n=NAME '(' params=parameters ')' a=['->' z=annotation { z }] ':' b=block {
+        _Py_AsyncFunctionDef(((expr_ty) n)->v.Name.id, params, b, NULL, a, NULL, EXTRA) }
+    | s=ASYNC 'def' n=NAME '(' ')' a=['->' z=annotation { z }] ':' b=block {
+        _Py_AsyncFunctionDef(((expr_ty) n)->v.Name.id, empty_arguments(p), b, NULL, a, NULL, EXTRA) }
+    | s='def' n=NAME '(' params=parameters ')' a=['->' z=annotation { z }] ':' b=block {
+        _Py_FunctionDef(((expr_ty) n)->v.Name.id, params, b, NULL, a, NULL, EXTRA) }
+    | s='def' n=NAME '(' ')' a=['->' z=annotation { z }] ':' b=block {
+        _Py_FunctionDef(((expr_ty) n)->v.Name.id, empty_arguments(p), b, NULL, a, NULL, EXTRA) }
```
gvanrossum: I suspect we can do some collapsing here too. The last two could share an action:
```
    'def' n=NAME '(' params=[parameters] ')' a=['->' z=annotation { z }] ':' b=block {
        _Py_FunctionDef(((expr_ty) n)->v.Name.id, params ? params : empty_arguments(p), b, NULL, a, NULL, EXTRA) }
```

pegen/c_generator.py
```diff
@@ -276,6 +276,14 @@ def _handle_default_rule_body(self, node: Rule, rhs: Rhs, result_type: str) -> N
                 with self.indent():
                     self.print("return res;")
             self.print("int mark = p->mark;")
+            self.print("if (p->mark == p->fill) {")
+            with self.indent():
+                self.print("if (fill_token(p) < 0) return NULL;")
+            self.print("}")
```
gvanrossum: You could just combine the two conditions:
```
if (p->mark == p->fill && fill_token(p) < 0)
    return NULL;
```

pegen/c_generator.py
```diff
@@ -399,6 +415,16 @@ def visit_Alt(
                     self.visit(item, names=names)
             self.print(") {")
             with self.indent():
+                self.print("assert(p->mark > 0);")
```
gvanrossum: You can move this assert into `get_last_nonnwhitespace_token()`.

pegen/pegen.c
```diff
@@ -262,6 +262,21 @@ expect_token(Parser *p, int type)
...
 }
 
+Token *
+get_last_nonnwhitespace_token(Parser *p)
+{
+    int i = 1;
+    Token *token = NULL;
+    while (i <= p->mark) {
+        token = p->tokens[p->mark-i];
+        if (token->type != 0 && (token->type < 4 || token->type > 6)) {
```
gvanrossum: Don't we have names for `0`, `4` and `6`?

pegen/pegen.c
```diff
@@ -262,6 +262,21 @@ expect_token(Parser *p, int type)
     return t;
 }
 
+Token *
+get_last_nonnwhitespace_token(Parser *p)
+{
+    int i = 1;
+    Token *token = NULL;
+    while (i <= p->mark) {
```
gvanrossum: Seems this can be formulated more readably using a `for` loop, e.g.
```
for (int m = p->mark - 1; m >= 0; m--) {
    token = p->tokens[m];
   ...
}
```

pegen/pegen.c
```diff
@@ -821,20 +840,32 @@ construct_assign_target(Parser *p, expr_ty node)
...
                 //TODO: We need to return a dummy here because we don't have a way to correctly
                 // buble up exceptions for now.
                return _Py_Name(_create_dummy_identifier(p),
                             Store,
-                            EXTRA_EXPR(node, node));
+                            node->lineno,
+                            node->col_offset,
+                            node->end_lineno,
+                            node->end_col_offset,
+                            p->arena);
```
gvanrossum: Seems maybe you could keep `EXTRA_EXPR` around just so it can be called by various helpers in pegen.c?

gvanrossum:
Marked it WIP because it has to wait until 3.8.1 has landed.

lysnikolaou:
I think, it would be better for the clearing of the grammar clutter to go into a new PR, as soon as we have merged this one. Otherwise this is gonna be very big and I'm concerned we might miss something.

Do you have an idea on how long it takes Travis to support new versions?

pablogsal:
>Do you have an idea on how long it takes Travis to support new versions?

Another possibility is using 'nightly' on Travis, although some dependencies may fail due to the lack of wheels (therefore compilation is needed and we would need the build dependencies) or other incompatibilities.

lysnikolaou:
> > Do you have an idea on how long it takes Travis to support new versions?
> 
> Another possibility is using 'nightly' on Travis, although some dependencies may fail due to the lack of wheels (therefore compilation is needed and we would need the build dependencies) or other incompatibilities.
> 
> We can certainly try.

We could try that. The release schedule specifies that 3.8.1 is going to be released tomorrow though, do you know if that's going to happen? If it is, then I think it's worth waiting.

PS I'm not that familiar with how `.travis.yml` should look, so I don't feel that comfortable making the changes, since it could break things.

pablogsal:
> We could try that. The release schedule specifies that 3.8.1 is going to be released tomorrow though, do you know if that's going to happen? If it is, then I think it's worth waiting.

What I am not sure is how much time takes travis to pick up the changes. If this is not urgent or blocking we could wait.



> PS I'm not that familiar with how `.travis.yml` should look, so I don't feel that comfortable making the changes, since it could break things.

I can do the appropriate changes to try it out if we need to, but as you said, I propose to wait a bit to see how much time does it take for them to make 3.8.1 available.

lysnikolaou:
Could we also use 3.8-dev?



> What I am not sure is how much time takes travis to pick up the changes.

It seems, they also rely on `pyenv` making 3.8.1 available, which means we could open a PR for `pyenv` to do it, as soons as 3.8.1 is released. It seems like it is not that hard to do.

pablogsal:
> Could we also use 3.8-dev?
Yes, and actually would be a better idea than using 'nightly'.

> It seems, they also rely on `pyenv` making 3.8.1 available, which means we could open a PR for `pyenv` to do it, as soons as 3.8.1 is released. It seems like it is not that hard to do.

Ok, what do you think if we start using 3.8-dev meanwhile we wait for pyenv? We did a similar thing when mypy could not handle PEP 572 syntax.

lysnikolaou:
> Ok, what do you think if we start using 3.8-dev meanwhile we wait for pyenv? We did a similar thing when mypy could not handle PEP 572 syntax.

That seems like the best solution to me.

pablogsal:
> That seems like the best solution to me.

Created https://github.com/gvanrossum/pegen/pull/125

pablogsal:
There are some simple merge conflicts around the `assignment` rule. We need to solve them first for Travis to run.

lysnikolaou:
> There are some simple merge conflicts around the assignment rule. We need to solve them first for Travis to run.

I also need to update the new refactorings around `lambda` for it to pass the tests.

pegen/c_generator.py
```diff
@@ -276,6 +276,14 @@ def _handle_default_rule_body(self, node: Rule, rhs: Rhs, result_type: str) -> N
                 with self.indent():
                     self.print("return res;")
             self.print("int mark = p->mark;")
+            self.print("if (p->mark == p->fill && fill_token(p) < 0) {")
```
pablogsal: Suggestion: Could we factor this into a method? This way we can easily identify
in the future what this new chunk is for and also it seems very similar (if not the same)
with the new addition in `_handle_loop_rule_body` and part of `visit_Alt`.

lysnikolaou: What about two methods? One for lineno/col_offset and one for end_lineno/end_col_offset?

pablogsal: That would be fine, although I think I prefer only as they are still very related. We can add maybe some comments before the lineno/col_offset  and the end_lineno/end_col_offset part. Being said that, I don't feel strongly either way :)

lysnikolaou: Would what I've done in 7a4f0f4 work?

pablogsal: > Would what I've done in [7a4f0f4](https://github.com/gvanrossum/pegen/commit/7a4f0f4d28246e70182d1e6460e45570327e0637) work?

Yeah, that is fantastic. Thanks!

pegen/c_generator.py
```diff
@@ -238,6 +238,27 @@ def generate(self, filename: str) -> None:
         mode = int(self.rules["start"].type == "mod_ty")
         self.print(EXTENSION_SUFFIX.rstrip("\n") % dict(mode=mode))
 
+    def _set_up_token_metadata_extraction(self, is_start: bool) -> None:
```
gvanrossum: This should be two separate methods.

PR #125, opened 2019-12-15
Use the 3.8-dev branch in Travis

pablogsal:
Meanwhile we wait for the 3.8.1 version to be available in travis,
use the development branch build to pick up the latest AST bugfixes.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,5 @@
 language: python
-python: 3.8
+python: 3.8-dev
 
 notifications:
   on_success: change


lysnikolaou:
Not sure about this, but I feel that this can be merged despite the `pytest` failures (that occur due to the lasted AST bugfixes), as these will be fixed after #121.

pablogsal:
As expected, the `if` rules are failing

> Not sure about this, but I feel that this can be merged despite the `pytest` failures (that occur due to the lasted AST bugfixes), as these will be fixed after #121.

In that case, I feel more comfortable if you include this change in #121

lysnikolaou:
> In that case, I feel more comfortable if you include this change in #121

Good idea! Let's do that.

PR #132, opened 2019-12-18
Create a more general way to set context in AST nodes

lysnikolaou:


test/test_data/with_tuple_target.py
```diff
@@ -0,0 +1,3 @@
+with a as (x, y):
```
pablogsal: I will add more tests that check the recursive case. For example:

```
with a as ((x,y), z):  pass
```

lysnikolaou: I was thinking that maybe `del_tuple.py` was enough, but still better to be sure.

pablogsal: Will do a complete review tomorrow! Today is late :)

lysnikolaou: Agreed!

pablogsal: > I was thinking that maybe `del_tuple.py` was enough, but still better safe than sorry.

Indeed, ideally we should have test cases for all possibilities to avoid regression due to potential refactors.

Issue #127, opened 2019-12-15
Simpy refactorings: F-Strings


lysnikolaou:
In the atom rule, the `STRING+` alternative still needs to be refactored.

gvanrossum:
I will create a separate issue that clearly says we still need to do f-strings.

PR #133, opened 2019-12-28
Support STRING+ atoms and add support multiple string types (but still no f-strings)

pablogsal:
Supporting f-strings will still require a considerable amount of code for the f-string parser so I will do this in a different PR as this one is already considerable big.

### DIFF
--- a/pegen/pegen.c
+++ b/pegen/pegen.c
@@ -2,9 +2,9 @@
 #include "pegen.h"
 #include "v38tokenizer.h"
 
-static PyObject *
-_create_dummy_identifier(Parser *p) {
-    PyObject *id = PyUnicode_FromStringAndSize("", 0);
+static inline PyObject *
+new_identifier(Parser *p, char* identifier) {
+    PyObject *id = PyUnicode_FromString(identifier);
     if (id == NULL) {
         return NULL;
     }
@@ -13,6 +13,11 @@ _create_dummy_identifier(Parser *p) {
         return NULL;
     }
     return id;
+}
+
+static PyObject *
+_create_dummy_identifier(Parser *p) {
+    return new_identifier(p, "");
 }
 
 static inline Py_ssize_t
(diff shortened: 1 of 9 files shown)


pegen/pegen.c
```diff
@@ -1295,3 +1301,350 @@ seq_delete_starred_exprs(Parser *p, asdl_seq *kwargs)
     }
     return new_seq;
 }
+
+
+//// STRING HANDLING FUNCTIONS ////
```
gvanrossum: Might note that these are all copied from Python/ast.c with slight modifications to compensate for the use of `Parser *p` instead of `struct compiling *c`. (And what else? I noticed some functions lacking a `node *n` argument as well.)

pegen/pegen.c
```diff
@@ -1295,3 +1301,350 @@ seq_delete_starred_exprs(Parser *p, asdl_seq *kwargs)
...
+        if (*rawmode)
+            *result = PyUnicode_DecodeUTF8Stateful(s, len, NULL, NULL);
+        else
+            *result = decode_unicode_with_escapes(p, s, len);
+    }
+    return *result == NULL ? -1 : 0;
+}
+
+expr_ty
+concatenate_strings(Parser *p, asdl_seq *strings)
```
gvanrossum: Except it looks like this function is new, right?

pablogsal: Yeah, this is the only function that does not have an equivalent in `Python/ast.c`.

lysnikolaou:
Also wouldn't it be better to wait on closing #127 until f-strings are also supported?

test/test_data/multipart_string.py
```diff
@@ -0,0 +1 @@
+"string1" "string2" "string3"
```
lysnikolaou: These should be included in the newly created `TEST_SOURCES` array from #134.

pablogsal: Thanks! I rebased to pick up the latest changes and forgot about moving these

pablogsal:
I have rebased to fix the merge conflicts but I cannot land myself because Travis fails after #138 as the fix for the col offsets is still not on Travis.

@gvanrossum Could you land this? Or should we wait until Travis picks up the latest changes?

gvanrossum:
Done. Though if it takes more than a day I recommend commenting out the offending tests.

lysnikolaou:
This just broke all the tests though due to a [typo in `pegen.c`](https://github.com/gvanrossum/pegen/blob/3be531ca10aeb94606a53be104f7e9d28a82e168/pegen/pegen.c#L1570). If this is removed, then tests are okay again. Should I just push a fix to master?

gvanrossum:
Whoops! Go ahead and push a fix.

lysnikolaou:
It seems, I don't have permissions. I'm getting some Travis error: `Required status check "Travis CI - Pull Request" is expected.`. I guess, it'd be better that you do it then, @gvanrossum.

gvanrossum:
I'd be happy to, but can you create a PR? I have no idea what the fix should be (I don't even know which tests are failing).

lysnikolaou:
Sure!

PR #134, opened 2019-12-28
Move all test cases to the test_ast_generation file

pablogsal:
Maintaining individual files for all test cases can be a bit cumbersome,
especially when trying to compare test cases among them and see what is
covered. If desired, multiline-strings can be used still to split test
cases visually in lines and avoid "\n" characters.

I find this structure more easy to work with but If you think that having
per-file test cases is still better, feel free to just close the PR :)


test/test_ast_generation.py
```diff
@@ -8,11 +8,122 @@
...
-)
+TEST_CASES = [
+    ('annotated_assignment', 'x: int = 42\n'),
+    ('annotated_assignment_with_parens', '(paren): int = 3+2\n'),
+    ('annotated_assignment_with_yield', 'x: int = yield 42\n'),
+    ('annotated_no_assignment', 'x: int\n'),
+    ('annotation_with_parens', '(parens): int\n'),
+    ('assert', 'assert a\n'),
+    ('assert_message', 'assert a, b\n'),
+    ('asyncfor', 'async for i in a:\n    pass\n'),
```
gvanrossum: For multi-line files I'd like to see the structure of the file more clearly here, e.g. using a triple-quoted string (with some utility to strip indents) or just string concatenation, e.g.
```
    ('asyncfor',
        'async for i in a:\n'
        '    pass\n'),
```
(Or perhaps I'd like it even better if the \n wasn't present -- then you could make it a list of lines.)

pablogsal: In commit 978bdf7 I have switched to multi-line strings (it reads better IMHO) and I have added preprocessing of the test cases that de-dents the string cases and also allows the possibility of using lists of strings as test cases (in which case they will be joined by newlines).

test/test_ast_generation.py
```diff
@@ -1,129 +1,299 @@
...
-    ('yield_expr', 'yield a\n'),
-    ('yield_from', 'yield from a\n'),
+    ('annotated_assignment', 'x: int = 42'),
+    ('annotated_assignment_with_parens', '(paren): int = 3+2'),
+    ('annotated_assignment_with_yield', 'x: int = yield 42'),
+    ('annotated_no_assignment', 'x: int'),
+    ('annotation_with_parens', '(parens): int'),
+    ('assert', 'assert a'),
+    ('assert_message', 'assert a, b'),
+    ('asyncfor', '''\
```
gvanrossum: Can you do something about the ugly `\` too here?

pablogsal: Done in 81c5b34

pablogsal:
> Okay, but you're going to have to exclude this from black by editing pyproject.toml, else `make black` will uglify it.

We could use `# fmt: off` and `# fmt:`  around the test cases so the rest of file can still be formated

gvanrossum:
Works for me!

lysnikolaou:
Nice one!

pablogsal:
> @pablogsal Could you also move the test cases that were added in #132?

I have rebased and moved the new tests in b47cc73

Issue #126, opened 2019-12-15
Simpy refactorings: primary


lysnikolaou:

lysnikolaou:
It seems there is something off with the line/col info of a generator expression inside a `Call`. Opened an issue [here](https://bugs.python.org/issue39235).

PR #135, opened 2020-01-05
Refactor rules for class_def and arguments

### LINKED ISSUE #122: Simpy refactorings: class_def and arguments

lysnikolaou:
Closes #122.

### DIFF
--- a/data/simpy.gram
+++ b/data/simpy.gram
@@ -184,7 +184,7 @@ annotation[expr_ty]: expression
 
 decorators[asdl_seq*]: a=('@' f=factor NEWLINE { f })+ { a }
 
-class_def:
+class_def[stmt_ty]:
     | a=decorators b=class_def_raw { class_def_decorators(p, a, b) }
     | class_def_raw
 class_def_raw[stmt_ty]:
@@ -390,13 +390,12 @@ args[expr_ty]:
                           seq_extract_starred_exprs(p, a),
                           seq_delete_starred_exprs(p, a),
                           EXTRA) }
-    | a=posarg b=[',' c=args { c }] {
+    | a=named_expression b=[',' c=args { c }] {
         _Py_Call(CONSTRUCTOR(p),
                  (b) ? seq_insert_in_front(p, a, ((expr_ty) b)->v.Call.args) : singleton_seq(p, a),
                  (b) ? ((expr_ty) b)->v.Call.keywords : NULL,
                  EXTRA) }
 kwargs[asdl_seq*]: a=','.kwarg+ { a }
-posarg[expr_ty]: starred_expression | named_expression
 starred_expression[expr_ty]:
     | '*' a=expression { _Py_Starred(a, Load, EXTRA) }
 kwarg[KeywordOrStarred*]:
--- a/data/simpy.gram
+++ b/data/simpy.gram
@@ -184,7 +184,15 @@ annotation[expr_ty]: expression
 
 decorators[asdl_seq*]: a=('@' f=factor NEWLINE { f })+ { a }
 
-class_def: [decorators] 'class' NAME ['(' [arguments] ')'] ':' block
+class_def:
+    | a=decorators b=class_def_raw { class_def_decorators(p, a, b) }
+    | class_def_raw
+class_def_raw[stmt_ty]:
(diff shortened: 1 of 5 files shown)


test/test_ast_generation.py
```diff
@@ -31,6 +31,37 @@
...
+    ('class_def_simple',
+     '''
+        class C:
+            pass
+     '''),
+    ('class_def_starred_in_kwargs',
+     '''
+        class C(A, x=2, *[B, C], y=3):
+            pass
+     '''),
```
pablogsal: I think we are missing a case with *args and **kwargs at the same time, which is valid in current Python:

```
>>> class A(x,y,*something,**more_something):   pass
```

lysnikolaou: Oh yes, thats right! Another problem is that this currently allows generator expressions in the arguments, which is not valid Python, if I'm not mistaken.

lysnikolaou:
@pablogsal @gvanrossum Is it okay if I push a PR that builds up on this for issues #126 and #136?

gvanrossum:
Yeah, as long as you’re not afraid of some merge conflicts.
-- 
--Guido (mobile)

data/simpy.gram
```diff
@@ -365,11 +373,37 @@ yield_expr[expr_ty]:
...
+    | a=starred_expression b=[',' c=args { c }] {
+        _Py_Call(CONSTRUCTOR(p),
+                 (b) ? seq_insert_in_front(p, a, ((expr_ty) b)->v.Call.args) : singleton_seq(p, a),
+                 (b) ? ((expr_ty) b)->v.Call.keywords : NULL,
+                 EXTRA) }
+    | a=kwargs { _Py_Call(CONSTRUCTOR(p),
+                          seq_extract_starred_exprs(p, a),
+                          seq_delete_starred_exprs(p, a),
+                          EXTRA) }
+    | a=posarg b=[',' c=args { c }] {
```
gvanrossum: Looks like we can get rid of `posarg` and instead write `a=named_expression` here -- the case where `posarg` matches `starred_expression` is already handled by the first alternative.

(Things are still weird -- one confusing thing is that `starred_expression` *must* start with `*`, while `named_expression` may or may not contain a `:=` operator. But I don't think we can simplify further, since the resolution of "ambiguities" is quite subtle for this rule.)

docs/helpers.md
```diff
@@ -83,6 +83,14 @@ Used to encapsulate the value of an operator_ty enum value.
 
 - kind: operator_ty, The augmented assignment operator
 
+#### KeywordOrStarred
+
+This is needed by the `arguments` rule, because a starred expression can appear
+after the first keyword argument has been parsed.
+
+- elem: `void *`, The keyword argument or starred expression itself
+- is_keyword: `int` (`bool`), Flag if elem is a keyword argument or a starred expression
```
gvanrossum: Must be 0 for `*args`, 1 for `kw=expr` and `**kwargs`.

docs/helpers.md
```diff
@@ -152,3 +160,18 @@ Constructs an empty `arguments_ty` object, that gets used when a function accept
...
 Creates an `AugOperator` encapsulating the operator type provided in *kind*.
+
+###### `stmt_ty function_def_decorators(Parser *p, asdl_seq *decorators, stmt_ty function_def)`
+Construct a `FunctionDef` equivalent to `function_def`, but with `decorators`.
+
+###### `stmt_ty class_def_decorators(Parser *p, asdl_seq *decorators, stmt_ty class_def)`
+Construct a `ClassDef` equivalent to `class_def`, but with `decorators`.
+
+###### `int _seq_number_of_starred_exprs(asdl_seq *seq)`
+Get the number of starred expressions in an `asdl_seq*` of `KeywordOrStarred*`s.
```
gvanrossum: Does this need to be documented? It's not public.

pegen/pegen.c
```diff
@@ -1204,3 +1205,93 @@ function_def_decorators(Parser *p, asdl_seq *decorators, stmt_ty function_def)
...
+    if (!a) {
+        return NULL;
+    }
+    a->element = element;
+    a->is_keyword = is_keyword;
+    return a;
+}
+
+/* Get the number of starred expressions in an asdl_seq* of KeywordOrStarred*s */
+int
```
gvanrossum: `static`

PR #137, opened 2020-01-05
Allow the C parser to make use of metas for header/footer inclusion

pablogsal:
This capability was missing from the C generator.


pegen/c_generator.py
```diff
@@ -217,7 +217,12 @@ def out_of_memory_goto(
 
     def generate(self, filename: str) -> None:
         self.collect_todo()
-        self.print(EXTENSION_PREFIX.format(filename=filename))
+        header = self.grammar.metas.get("header", EXTENSION_PREFIX)
+        if header is not None:
```
gvanrossum: Why not `if header:` here?

pegen/c_generator.py
```diff
@@ -217,7 +217,12 @@ def out_of_memory_goto(
 
     def generate(self, filename: str) -> None:
         self.collect_todo()
-        self.print(EXTENSION_PREFIX.format(filename=filename))
```
gvanrossum: Are you sure you want to allow overriding the `// @generated by pegen.py from {filename}` line?

pegen/c_generator.py
```diff
@@ -236,7 +241,9 @@ def generate(self, filename: str) -> None:
                 self.print()
                 self.visit(rule)
         mode = int(self.rules["start"].type == "mod_ty")
-        self.print(EXTENSION_SUFFIX.rstrip("\n") % dict(mode=mode))
```
gvanrossum: ISTM that this encourages people to copy the entire body of `EXTENSION_SUFFIX` in order to change the module name. Perhaps we can add a variable to override just that, so fewer people will have to set `@trailer`?

test/test_c_parser.py
```diff
@@ -303,3 +305,31 @@ def test_syntax_error_for_file(text: str, tmp_path: PurePath) -> None:
         tb = traceback.format_exc()
     assert 'some_file.py", line 1' in tb
     assert f"{text}\n        ^" in tb
+
+
+def test_metas(tmp_path: PurePath) -> None:
+    grammar_source = f"""
```
gvanrossum: This is pretty obfuscated, using an f-string containing more triple-quoted strings and backslashes. I have no idea of the purpose of the code you add in the headers and trailer.

test/test_c_parser.py
```diff
@@ -303,3 +305,31 @@ def test_syntax_error_for_file(text: str, tmp_path: PurePath) -> None:
...
+            return x;
+        }}
+        {textwrap.indent(EXTENSION_SUFFIX, "    ")}
+    '''
+    start: expr+ NEWLINE? ENDMARKER
+    expr: x=NAME {{identity(x)}}
+    """
+    grammar = parse_string(grammar_source, GrammarParser)
+    extension = generate_parser_c_extension(grammar, tmp_path)
+    extension.parse_string("a b c")
```
gvanrossum: Why not write a test that tests exactly what is being generated, rather than trying to compile and run it? If necessary, allow better dependency injection to allow such a test to be written.

pablogsal:
> What use case led to this? It all seems a bit complicated.

In the C generator case, allowing to use a subheader to include extra headers from the grammar file. For the Python one, there is an example of use in the metagrammar.

test/test_c_parser.py
pablogsal: > This is pretty obfuscated, using an f-string containing more triple-quoted strings and backslashes. I have no idea of the purpose of the code you add in the headers and trailer.

The idea was that the code can only be compiled and executed if the three pieces are included in the final result (so it checks that header/subheader and trailer work instead of just checking for one).

test/test_c_parser.py
pablogsal: > Why not write a test that tests exactly what is being generated, rather than trying to compile and run it?

Originally, I wanted to reuse as much machinery as possible, but I agree that testing the contents of the generated would make the test simpler.

pegen/c_generator.py
pablogsal: Good idea!

pegen/c_generator.py
```diff
@@ -25,8 +25,6 @@
 from pegen.tokenizer import exact_token_types
 
 EXTENSION_PREFIX = """\
-// @generated by pegen.py from {filename}
-
 #include "pegen.h"
 """
 EXTENSION_SUFFIX = """
```
gvanrossum: Get rid of the TODO comment one line below.

pegen/c_generator.py
```diff
@@ -241,9 +240,10 @@ def generate(self, filename: str) -> None:
                 self.print()
                 self.visit(rule)
         mode = int(self.rules["start"].type == "mod_ty")
+        modname = self.grammar.metas.get("modulename", "parse")
```
gvanrossum: Nit: spell `modulename` in full.

PR #138, opened 2020-01-06
Refactor primary and slices rules

### LINKED ISSUE #126: Simpy refactorings: primary

### LINKED ISSUE #136: Simpy refactorings: slices

lysnikolaou:
Marking this as WIP until the [genexp issue](https://bugs.python.org/issue39235) on bpo is resolved.

Closes #126 and closes #136.

### DIFF
--- a/data/simpy.gram
+++ b/data/simpy.gram
@@ -324,10 +324,22 @@ power[expr_ty]:
     | primary 
 primary[expr_ty]:
     | !'await' a=atom !'.' !'(' !'[' { a }
-    | [AWAIT] atom ('.' NAME | '[' slices ']' | '(' [arguments] ')')*
+    | AWAIT a=primary_raw { _Py_Await(a, EXTRA) }
+    | primary_raw
+primary_raw[expr_ty]:
+    | a=primary_raw '.' b=NAME { _Py_Attribute(a, ((expr_ty) b)->v.Name.id, Load, EXTRA) }
+    | a=primary_raw '[' b=slices ']' { _Py_Subscript(a, b, Load, EXTRA) }
+    | a=primary_raw '(' b=[arguments] ')' {
+        _Py_Call(a,
+                 (b) ? ((expr_ty) b)->v.Call.args : NULL,
+                 (b) ? ((expr_ty) b)->v.Call.keywords : NULL,
+                 EXTRA) }
+    | atom
 
-slices: ','.slice+ [',']
-slice: [expression] ':' [expression] [':' [expression]] | expression
+slices[slice_ty]: a=','.slice+ [','] { (asdl_seq_LEN(a) == 1) ? asdl_seq_GET(a, 0) : _Py_ExtSlice(a, p->arena) }
+slice[slice_ty]:
+    | a=[expression] ':' b=[expression] c=[':' d=[expression] { d }] { _Py_Slice(a, b, c, p->arena) }
+    | a=expression { _Py_Index(a, p->arena) }
 # STRING+'s output is just a hack for now 
 atom[expr_ty]:
     | list
--- a/data/simpy.gram
+++ b/data/simpy.gram
@@ -328,6 +328,7 @@ primary[expr_ty]:
     | primary_raw
 primary_raw[expr_ty]:
     | a=primary_raw '.' b=NAME { _Py_Attribute(a, ((expr_ty) b)->v.Name.id, Load, EXTRA) }
+    | a=primary_raw '[' b=slice_expressions ']' { _Py_Subscript(a, b, Load, EXTRA) }
     | a=primary_raw '[' b=slices ']' { _Py_Subscript(a, b, Load, EXTRA) }
     | a=primary_raw '(' b=[arguments] ')' {
         _Py_Call(a,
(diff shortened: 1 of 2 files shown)


data/simpy.gram
```diff
@@ -324,10 +324,22 @@ power[expr_ty]:
...
+                 (b) ? ((expr_ty) b)->v.Call.keywords : NULL,
+                 EXTRA) }
+    | atom
 
-slices: ','.slice+ [',']
-slice: [expression] ':' [expression] [':' [expression]] | expression
+slices[slice_ty]: a=','.slice+ [','] { (asdl_seq_LEN(a) == 1) ? asdl_seq_GET(a, 0) : _Py_ExtSlice(a, p->arena) }
+slice[slice_ty]:
+    | a=[expression] ':' b=[expression] c=[':' d=[expression] { d }] { _Py_Slice(a, b, c, p->arena) }
+    | a=expression { _Py_Index(a, p->arena) }
```
gvanrossum: Seems the logic in `ast.c` (in `ast_for_trailer()`) is different. If there are multiple slices but none of them have colons, it uses `Index(Tuple(...))` rather than `ExtSlice(...)`. Using the power of PEG and backtracking we could code this by using an extra alternative in `primary_raw`:
```
    | a=primary_raw '[' b=slice_expressions ']' { ... }
    | a=primary_raw '[' b=slices ']' { ... }
```
so that backtracking will happen if there are any colons. Then we need
```
slice_expressions: ','.expression+ [','] { ... }
```
Alternatively we could write a helper function to implement that logic (which may result in less generated code but more code to maintain, in a sense).

lysnikolaou: Oh yeah, for some reason I didn't check to see what `ast.c` does. Sorry!

data/simpy.gram
```diff
@@ -336,7 +337,11 @@ primary_raw[expr_ty]:
                  EXTRA) }
     | atom
 
-slices[slice_ty]: a=','.slice+ [','] { (asdl_seq_LEN(a) == 1) ? asdl_seq_GET(a, 0) : _Py_ExtSlice(a, p->arena) }
+slice_expressions[slice_ty]: a=','.expression+ [','] {
+    (asdl_seq_LEN(a) == 1) ? _Py_Index(asdl_seq_GET(a, 0), p->arena)
+                           : _Py_Index(_Py_Tuple(a, Load, EXTRA), p->arena) }
+slices[slice_ty]: a=','.slice+ [','] {
+    (asdl_seq_LEN(a) == 1) ? asdl_seq_GET(a, 0) : _Py_ExtSlice(a, p->arena) }
```
gvanrossum: Ditto here -- you could use
```
    | a=slice !',' { a }
    | a=','.slice+ { _Py_ExtSlice(a, p->arena) }
```

lysnikolaou:
@gvanrossum Is this the way?

data/simpy.gram
```diff
@@ -324,10 +324,28 @@ power[expr_ty]:
     | primary 
 primary[expr_ty]:
     | !'await' a=atom !'.' !'(' !'[' { a }
```
gvanrossum: I think we don't need this one any more. It was a shortcut from the time when most primaries didn't generate an AST at all. (Though it possibly speeds up parsing?)

data/simpy.gram
```diff
@@ -324,10 +324,28 @@ power[expr_ty]:
     | primary 
 primary[expr_ty]:
```
gvanrossum: I wonder if we could rename this to `await_primary` and then rename `primary_raw` to just `primary`.

data/simpy.gram
```diff
@@ -324,10 +324,28 @@ power[expr_ty]:
...
-
-slices: ','.slice+ [',']
-slice: [expression] ':' [expression] [':' [expression]] | expression
+    | AWAIT a=primary_raw { _Py_Await(a, EXTRA) }
+    | primary_raw
+primary_raw[expr_ty]:
+    | a=primary_raw '.' b=NAME { _Py_Attribute(a, ((expr_ty) b)->v.Name.id, Load, EXTRA) }
+    | a=primary_raw '[' b=expression ']' { _Py_Subscript(a, _Py_Index(b, p->arena), Load, EXTRA) }
+    | a=primary_raw '[' b=slice_expressions ']' { _Py_Subscript(a, b, Load, EXTRA) }
+    | a=primary_raw '[' b=slices ']' { _Py_Subscript(a, b, Load, EXTRA) }
```
gvanrossum: An alternative way to factor this (up to you) is to introduce a sub-rule ("slicing"?) that represents the `[` ... `]` part.

data/simpy.gram
lysnikolaou: Let's go with the simplified version for now and care about speed when we have a complete grammar!

data/simpy.gram
lysnikolaou: Wouldn't this need a helper type, since `expression` returns an `expr_ty`, but `slice_expressions` and `slices` return `slice_ty`? If that's so, I'd leave it as is.

gvanrossum: I was thinking of this:
```
primary[expr_ty]:
    ...
    | a=primary b=slicing { _Py_Subscript(a, b, Load, EXTRA) }
    ...
slicing[slice_ty]:
    | '[' b=expression ']' { _Py_Index(b, p->arena) }
    | '[' b=slice_expressions ']' { b }
    | '[' b=slices ']' { b }
```

lysnikolaou:
@gvanrossum Yeah, Serhiy hasn't responded yet, so we need to wait.

On another note, could you please briefly explain why in this case
```
slicing[slice_ty]:
    | '[' b=expression ']' { _Py_Index(b, p->arena) }
    | '[' b=(slice_expressions | slices) ']' { b }
```
won't work? I tried it out and found out that it's not possible to merge the two alternatives into one, but I can't really figure out why.

gvanrossum:
It’s because of the way backtracking works, and because a parenthesized group gets turned into an anonymous rule. The failure doesn’t get apparent until after this rule succeeds, with the first alternative, and retrying will just make it return the same result.

lysnikolaou:
> and retrying will just make it return the same result.

This gets me every time! Thanks!

gvanrossum:
I guess you will have to rework this so it works for `f(i for i in a)` now that https://bugs.python.org/issue39235 is fixed?

lysnikolaou:
Just did! Do we have an idea where Travis picks up the changes?

gvanrossum:
I have no idea when Travis picks up 3.8 branch changes. The info in https://docs.travis-ci.com/user/languages/python/#python-versions doesn't say much about this, just that `3.8-dev` exists.

I have verified that this passes all tests with the 3.8 HEAD, so you can land now. Thanks!

lysnikolaou:
I posted on the Travis community forum how ofter dev releases get built and the answer was about 24 hours, so it's not going to be much longer.

You can find the post [here](https://travis-ci.community/t/development-release-build-frequency/6804).

PR #139, opened 2020-01-09
Refactor targets rules

### LINKED ISSUE #123: Simpy refactorings: targets

lysnikolaou:
I'm pushing this PR, which uses the approach I was talking about on the issue before @gvanrossum suggested the other one with the two rules. This works correctly and I think that it is a bit more readable.

Closes #123.


lysnikolaou:
I found a serious bug on this. It rejects input, such as `a()().b`, because the closing parenthesis of the first call is not followed by either a `[` or a `.`. I guess, I'll have to work on it a bit longer.

lysnikolaou:
Superseded by #141.

PR #140, opened 2020-01-10
Typo fix in pegen.c

lysnikolaou:


lysnikolaou:
@gvanrossum I can't merge this due to Travis failing, so could you do it?

pablogsal:
Thanks for the catch and the fix, @lysnikolaou!

I am a bit surprised Travis didn't catch this (In https://travis-ci.com/gvanrossum/pegen/jobs/274043173 the only failing test is the generator expression one).

Should we explore adding a debug target to Travis that triggers the asserts?

lysnikolaou:
@pablogsal How would we do that? Do we need to change something in our Travis configuration?

pablogsal:
> @pablogsal How would we do that? Do we need to change something in our Travis configuration?

Opened https://github.com/gvanrossum/pegen/pull/143 for that :)

PR #141, opened 2020-01-11
Refactor targets rules

### LINKED ISSUE #123: Simpy refactorings: targets

lysnikolaou:
Closes #123.

### DIFF
--- a/data/simpy.gram
+++ b/data/simpy.gram
@@ -427,7 +427,23 @@ star_target: '*' bitwise_or | atom t_tail+ | a=star_atom t_tail* { a }
 star_atom: a=NAME { set_expr_context(p, a, Store) } | '(' [star_targets] ')' | '[' [star_targets] ']'
 
 targets[asdl_seq*]: a=','.target+ [','] { a }
-target[expr_ty]: atom t_tail+ | a=t_atom t_tail* { a }
+target[expr_ty]:
+    | a=non_terminal_target '.' b=NAME !tmore {
+        _Py_Attribute(a, ((expr_ty) b)->v.Name.id, Load, EXTRA) }
+    | a=non_terminal_target b=slicing !tmore { _Py_Subscript(a, b, Load, EXTRA) }
+    | t_atom
+non_terminal_target[expr_ty]:
+    | a=non_terminal_target '.' b=NAME &tmore {
+        _Py_Attribute(a, ((expr_ty) b)->v.Name.id, Load, EXTRA) }
+    | a=non_terminal_target b=slicing &tmore { _Py_Subscript(a, b, Load, EXTRA) }
+    | a=non_terminal_target b=genexp  &tmore { _Py_Call(a, singleton_seq(p, b), NULL, EXTRA) }
+    | a=non_terminal_target '(' b=[arguments] ')'  &tmore {
+        _Py_Call(a,
+                 (b) ? ((expr_ty) b)->v.Call.args : NULL,
+                 (b) ? ((expr_ty) b)->v.Call.keywords : NULL,
+                 EXTRA) }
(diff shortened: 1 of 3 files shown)


lysnikolaou:
There is the question, @gvanrossum mentioned in #123, if we wanna remove all the context functions in `pegen.c` and  just duplicate `targets` and have something like `del_targets` and `store_targets`.

gvanrossum:
> There is the question, @gvanrossum mentioned in #123, if we wanna remove all the context functions in `pegen.c` and just duplicate `targets` and have something like `del_targets` and `store_targets`.

Let's do that in a separate PR, if we feel like it.

data/simpy.gram
```diff
@@ -427,7 +427,21 @@ star_target: '*' bitwise_or | atom t_tail+ | a=star_atom t_tail* { a }
 star_atom: a=NAME { set_expr_context(p, a, Store) } | '(' [star_targets] ')' | '[' [star_targets] ']'
 
 targets[asdl_seq*]: a=','.target+ [','] { a }
-target[expr_ty]: atom t_tail+ | a=t_atom t_tail* { a }
+target[expr_ty]:
+    | a=non_terminal_target '.' b=NAME !tmore { _Py_Attribute(a, b->v.Name.id, Load, EXTRA) }
+    | a=non_terminal_target b=slicing !tmore { _Py_Subscript(a, b, Load, EXTRA) }
```
gvanrossum: Why not use `Store` here?

data/simpy.gram
```diff
@@ -427,7 +427,21 @@ star_target: '*' bitwise_or | atom t_tail+ | a=star_atom t_tail* { a }
...
+    | t_atom
+non_terminal_target[expr_ty]:
```
gvanrossum: While I agree that `target1` was a bad name showing lack of imagination, `non_terminal_target` feels too long. How about `t_primary`? Because it's exactly the same as `primary` except for the lookaheads, and we already have `t_atom` that's the target equivalent of `atom`.

data/simpy.gram
```diff
@@ -427,7 +427,21 @@ star_target: '*' bitwise_or | atom t_tail+ | a=star_atom t_tail* { a }
...
+    | a=non_terminal_target '.' b=NAME &tmore { _Py_Attribute(a, b->v.Name.id, Load, EXTRA) }
+    | a=non_terminal_target b=slicing &tmore { _Py_Subscript(a, b, Load, EXTRA) }
+    | a=non_terminal_target b=genexp  &tmore { _Py_Call(a, singleton_seq(p, b), NULL, EXTRA) }
+    | a=non_terminal_target '(' b=[arguments] ')'  &tmore {
+        _Py_Call(a,
+                 (b) ? ((expr_ty) b)->v.Call.args : NULL,
+                 (b) ? ((expr_ty) b)->v.Call.keywords : NULL,
+                 EXTRA) }
+    | a=atom &tmore { a }
+tmore: '(' | '[' | '.'
```
gvanrossum: Maybe rename to `t_lookahead` to clarify its purpose?

test/test_ast_generation.py
```diff
@@ -387,8 +393,25 @@ def f(*a, b):
...
+    ("del_call", "del a()"),
+    ("del_call_genexp", "del a(i for i in b)"),
+    ("del_subscript_call", "del a[b]()"),
+    ("del_attribute_call", "del a.b()"),
+    ("del_mixed_call", "del a[0].b().c.d()"),
+]
+
 # fmt: on
 
+def create_source_string(source: Any) -> str:
```
gvanrossum: I'd give this a different name, e.g. `cleanup_source()`.

PR #142, opened 2020-01-11
Generate type for NAME, STRING and NUMBER variables

lysnikolaou:
We can now avoid unnecessary casts in the grammar.

### DIFF
--- a/pegen/c_generator.py
+++ b/pegen/c_generator.py
@@ -492,6 +492,8 @@ class CParserGenerator(ParserGenerator, GrammarVisitor):
                     type = rule.type
             elif name.startswith("_loop") or name.startswith("_gather"):
                 type = "asdl_seq *"
+            elif name.startswith("name") or name.startswith("string") or name.startswith("number"):
+                type = "expr_ty"
         if node.name:
             name = node.name
         name = dedupe(name, names)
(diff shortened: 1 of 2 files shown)


pegen/c_generator.py
```diff
@@ -492,6 +492,8 @@ def add_var(self, node: NamedItem, names: List[str]) -> Tuple[str, Optional[str]
                     type = rule.type
             elif name.startswith("_loop") or name.startswith("_gather"):
                 type = "asdl_seq *"
```
gvanrossum: Please use exact comparisons (e.g. ‘name_var’).

lysnikolaou:
Done!

pegen/c_generator.py
```diff
@@ -492,7 +492,7 @@ def add_var(self, node: NamedItem, names: List[str]) -> Tuple[str, Optional[str]
...
-            elif name.startswith("name") or name.startswith("string") or name.startswith("number"):
+            elif name == "name_var" or name == "string_var" or name == "number_var":
```
gvanrossum: This is better written as `name in ("name_var", ...)`.

lysnikolaou: Right!

PR #143, opened 2020-01-12
Do not optimize away asserts in the CI

pablogsal:
This makes sure that assets are not optimized away (they are now because the extension inherits the config variables from the Python build in Travis and that one is not compiled in debug mode). This change will allow Travis to detect failures similar to the one in #140 as well as check that the asserts themselves do not fail.

### DIFF
--- a/pegen/__main__.py
+++ b/pegen/__main__.py
@@ -43,6 +43,9 @@ argparser.add_argument(
     help="Where to write the generated parser (default parse.py or parse.c)",
 )
 argparser.add_argument("filename", help="Grammar description")
+argparser.add_argument(
+    "--optimized", action="store_true", help="Compile the extension in optimized mode"
+)
 
 
 def main() -> None:
@@ -67,6 +70,7 @@ def main() -> None:
             verbose_tokenizer,
             verbose_parser,
             args.verbose,
+            keep_asserts_in_extension=False if args.optimized else True,
         )
     except Exception as err:
         if args.verbose:
--- a/pegen/build.py
+++ b/pegen/build.py
@@ -21,7 +21,10 @@ MOD_DIR = pathlib.Path(__file__)
 
 
 def compile_c_extension(
-    generated_source_path: str, build_dir: Optional[str] = None, verbose: bool = False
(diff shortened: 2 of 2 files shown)


pablogsal:
> I'd be more comfortable if this also set that flag when compiling locally. I just tried to test that this actually works by patching in this PR and then putting the `assert(const->kind...)` typo back, but it didn't fail.

Right, if you want it to work locally you would need to export the env variable:

```
export CFLAGS="$CFLAGS -UNDEBUG
```

> OTOH we probably don't want this flag for production use, since it slows down the parser (though I don't know by how much).

I had the same thought, that is why I was thinking on only activating it in the CI. On the other hand, what you would think about adding a new flag to `pytest` so you can do something like (modulo a better name for the flag)

```
python -m pytest tests --with-assertions
```

?

gvanrossum:
I think it should be on by default everywhere, but we need to have a way to turn it off if we ever make a setup.py to put on pypi, or for speed tests. Maybe the flag can be set in the helper function that builds the binary?

Also, I am a little concerned that speed is going down -- it's now about 3M bytes/sec for simpy_cpython, I believe it was 4M or maybe even 6M earlier. (There used to be a file with some early timings, but I can't find it right now.)

pablogsal:
> I think it should be on by default everywhere, but we need to have a way to turn it off if we ever make a setup.py to put on pypi, or for speed tests. Maybe the flag can be set in the helper function that builds the binary?

Good idea! Let me explore that and I will update this PR.

pablogsal:
@gvanrossum I have updated the PR so it keeps the asserts by default and to allow to pass down a `--optimized` version when calling `pegen` to deactivate the asserts (I called `optimized` so if we activate/deactivate more stuff in the future we don't need to add more flags). For example:

```
python -m pegen -q -c data/cprog.gram -o pegen/parse.c --compile-extension --optimized
```

pablogsal:
Here you have a PR with this patch and a typo in `pegen.c` so you can check that it indeed fails:

https://github.com/pablogsal/pegen/pull/2

Issue #129, opened 2019-12-16
Can we support named variables inside groups?


gvanrossum:
Suppose we have this rule:
```
rule: foo [bar [baz]] { Thing(bar, baz) }
```
This currently doesn't work because the scope of the bar and baz variable is just the smallest enclosing group. We can handle a simpler case just fine:
```
rule: foo b=[bar] { Thing(b, NULL) }
```
because the optional rule will return NULL if it isn't present. But extending this pattern to the first example would require returning a tuple, like this:
```
rule: foo t=[b1=bar b2=[baz] {tuple(b1, b2)}] { t ? Thing(t.b1, t.b2) : Thing(NULL, NULL) }
```
If we changed the scoping rules so that we could extract values directly from nested groups (as long as they physically appear in the alternative) we could write this directly as
```
rule: foo [b1=bar [b2=baz]] { Thing(b1, b2) }
```

gvanrossum:
Note that this will be hard, since it would require a totally different approach to how we generate code for nested groups. So I'm not expecting action. But it would still make for cleaner grammars (see e.g. `'raise'` in sympy.gram).

lysnikolaou:
Here's an approach, which can be implemented without giving up how we currently generate functions for groups.

In a nutshell: When we see a `Group` or an `Opt`, we recursively `collect_vars` for that as well. In the generated code for the parent `Alt` we pass pointers around to the `_tmp` function, which then propagate all their `NamedItem`s up to their parent through those pointers.

It turns out it's not really simple to implement, for me at least, but it's not that hard either. Could you give me some feedback on the idea? Do you think it's okay design-wise to do that? Should I go on with its implementation?

gvanrossum:
Yeah, that's a reasonable approach. It does sound like it'll cause a lot of complexity in the generated code, since we still want to support actions inside the group, and those actions must see simple local variables with the right names; so that means that we'll have to copy the values from those locals to the passed-in references upon every return.

Since we already have written the whole grammar without this feature, I propose to shelve the idea and focus on more important issues.

lysnikolaou:
Agreed.

PR #146, opened 2020-01-13
Add del_targets

lysnikolaou:
This deletes the only call to `set_seq_context` from the grammar, so I think it's actually useful.


lysnikolaou:
> You're up late. :-)

I was trying out something about #129(more on that tomorrow) and didn't realise it was this late! So much for today!

PR #148, opened 2020-01-14
Added grammar_grapher.py

bfoz:
This is an experiment in trying to visualize the overall structure of a grammar. The goal was to quickly get an idea of a grammar's complexity, and in particular, glean some insight into its "recursiveness". I'm not sure that I accomplished that goal, or that graphviz is the right tool for the job, but the resulting graphs are, at least, interesting and hopefully of some use.

I think with more refinement the output graphs can be made more useful (maybe with some colors and better labels, possibly some animation), but this seems like a good checkpoint for sharing what I have so far.

### DIFF
--- a/scripts/grammar_grapher.py
+++ b/scripts/grammar_grapher.py
@@ -1,0 +1,96 @@
+#!/usr/bin/env python3.8
+
+""" Convert a grammar into a dot-file suitable for use with GraphViz
+
+    For example:
+        Generate the GraphViz file:
+        # scripts/grammar_grapher.py data/simpy.gram > simpy.gv
+
+        Then generate the graph...
+
+        # twopi simpy.gv -Tpng > simpy_twopi.png
+
+        or
+
+        # dot simpy.gv -Tpng > simpy_dot.png
+
+        NOTE: The _dot_ and _twopi_ tools seem to produce the most useful results.
+              The _circo_ tool is the worst of the bunch. Don't even bother.
+"""
+
+import argparse
+import sys
+
+sys.path.insert(0, ".")
+
+from pegen.build import build_parser
+from pegen.grammar import Alt, Cut, Grammar, Group, Leaf, Lookahead, Rule, NameLeaf, NamedItem, Opt, Repeat, Rhs
+
+argparser = argparse.ArgumentParser(
+    prog="graph_grammar",
+    description="Graph a grammar tree",
+)
+argparser.add_argument("grammar_file", help="The grammar file to graph")
+
+def references_for_item(item):
+    if isinstance(item, Alt):
+        return [_ref for _item in item.items for _ref in references_for_item(_item)]
+    elif isinstance(item, Cut):
+        return []
+    elif isinstance(item, Group):
+        return references_for_item(item.rhs)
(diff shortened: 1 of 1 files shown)


gvanrossum:
This is cool. Unfortunately for the only grammar I really care about (`data/simpy.gram`), the resulting graph, at least when rendered with the default `dot` invocation, is pretty hard to read.
```
python3 scripts/grammar_grapher.py  data/simpy.gram  | dot  -T jpg >foo.jpg
open foo.jpg
```
foo

gvanrossum:
(And I just realized that that's basically what you yourself said. :-)

bfoz:
It is definitely hard to read. Harder than I was hoping for.

But, despite that, there's still some obvious clustering in the grammar. In particular, _bitwise_or_ stands out as being "popular", or at least well-connected. There's also a lot of self-recursion that's visible if you zoom in. Would it be useful to highlight those elements?

I'd like to think that the mess in the lower-left corner indicates a part of the grammar that could benefit from some TLC, but I'm not sure how that translates into anything actionable. At the very least, it suggests that all roads lead to the _NAME_ element.

From here I think I'll create an animation of the grammar's evolution as you work on it, just to see if that produces anything interesting. I'm also toying with the idea of making a graphical version of the graph_visualizer script to help with debugging.

In the end, I'm not sure what to make of the graphs either. Is messiness bad? Is clustering good? What do the more linear parts of the graph indicate? I really have no idea. It's a fun diversion, though.

gvanrossum:
Well, staring at the graph didn't cause me to reach any great insights, other than that `dot` isn't very good at visualizing graphs... :-(

Basically, grammars are expected to be recursive and there's nothing special about following the recursion paths. The various patterns are mostly due to patterns in the language (e.g. we have a lot of operator precedences) or ways that the grammar is written to be able to construct ASTs that are compatible with the existing CPython API. Bitwise or is special because it's the lowest precedence that's not a comparison or Boolean operator. And so on.

With the old pgen we used to have a tool that produced railroad diagrams of the grammar. These might be actually useful for users of the language. Another useful graph is the graph you get by only following left-recursive links (taking into account items that may have an empty expansion), since this graph is needed by the parser generator. But IIRC we currently don't actually have any non-obvious left-recursion (since we suspect there's a bug in the generator or in the runtime support for it, so we worked around it -- see https://github.com/gvanrossum/pegen/issues/123#issuecomment-573214363).

gvanrossum:
Questions for @lysnikolaou, @pablogsal -- should we land this anyway (it may be the starting point for something better)?

pablogsal:
> Questions for @lysnikolaou, @pablogsal -- should we land this anyway (it may be the starting point for something better)?

Although I agree that the output is confusing, is not much code and I agree that can be useful in the future. I am +1 to land this if everyone else agrees :)

gvanrossum:
OK, let's land it.

bfoz:
Thanks!

I'll look into making railroad diagrams to see if it's something `dot` can handle.

I like the idea of graphing the left-recursive spine. I'll see if there a way to get `dot` to treat those edges as a sort of trunk, with the rest of the grammar hanging off if it.

PR #153, opened 2020-01-15
Refactor visit_Alt in the C generator

pablogsal:
While I was working on some improvements on the C generator and realized that the code for visiting alternatives was getting very complex and there were some opportunities for some refactoring:

To disentangle the generation of code for alternatives, separate explicitly the options from the repetitions as they are handled differently and have different cleanups. Additionally, create functions
for emitting the default action or the one specified for clarity.

Better name suggestions for the functions are welcomed! :)

### DIFF
--- a/pegen/c_generator.py
+++ b/pegen/c_generator.py
@@ -451,7 +451,7 @@ class CParserGenerator(ParserGenerator, GrammarVisitor):
             self.print(f"goto done;")
         self.print("}")
 
-    def _handle_repetitions(
+    def _handle_alt_loop(
         self, node: Alt, is_gather: bool, rulename: Optional[str], names: List[str]
     ) -> None:
         # Condition of the main body of the alternative
@@ -492,7 +492,7 @@ class CParserGenerator(ParserGenerator, GrammarVisitor):
 
             names: List[str] = []
             if is_loop:
-                self._handle_repetitions(node, is_gather, rulename, names)
+                self._handle_alt_loop(node, is_gather, rulename, names)
             else:
                 self._handle_options(node, is_gather, names)
 
--- a/pegen/c_generator.py
+++ b/pegen/c_generator.py
@@ -435,7 +435,7 @@ class CParserGenerator(ParserGenerator, GrammarVisitor):
                 )
             self.print(f"res = {names[0]};")
 
-    def handle_alt_optional(self, node: Alt, is_gather: bool, names: List[str]) -> None:
+    def handle_alt_normal(self, node: Alt, is_gather: bool, names: List[str]) -> None:
         self.join_conditions(keyword="if", node=node, names=names)
         self.print("{")
         # We have parsed successfully all the conditions for the option.
@@ -494,7 +494,7 @@ class CParserGenerator(ParserGenerator, GrammarVisitor):
             if is_loop:
                 self.handle_alt_loop(node, is_gather, rulename, names)
             else:
-                self.handle_alt_optional(node, is_gather, names)
+                self.handle_alt_normal(node, is_gather, names)
 
             self.print("p->mark = mark;")
             if "cut_var" in names:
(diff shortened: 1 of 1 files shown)


pegen/c_generator.py
```diff
@@ -398,68 +398,104 @@ def visit_Rhs(
...
+            if node.action:
+                self._emit_action(node)
+            else:
+                self._emit_default_action(is_gather, names, node)
+
+            # As the current option has parsed correctly, do not continue with the rest.
+            self.print(f"goto done;")
+        self.print("}")
+
+    def _handle_repetitions(
```
lysnikolaou: I'd call it `_handle_loop`, since this is what we call it all throughout `c_generator.py`.

lysnikolaou:
Could the `simpy_cpython` failure be related to #154?

pablogsal:
> Could the `simpy_cpython` failure be related to #154?

I think it is because commit ba11136 in this PR was passing the CI and 93ea64f is just a rename. Maybe something changed in the nightly build of CPython.....

lysnikolaou:
Do you use a debug version of python locally?

pegen/c_generator.py
```diff
@@ -398,68 +398,104 @@ def visit_Rhs(
...
+
+    def _handle_alt_loop(
+        self, node: Alt, is_gather: bool, rulename: Optional[str], names: List[str]
+    ) -> None:
+        # Condition of the main body of the alternative
+        self._join_conditions(keyword="while", node=node, names=names)
+        self.print("{")
+        # We have parsed successfully one item!
+        with self.indent():
+            # Prepare to emmit the rule action and do so
```
gvanrossum: Emit

pablogsal:
> Why are these helpers named with a leading underscore?

I did it to distinguish "main flow" from helpers, but I agree that looks inconsistent.

> I’d rename ‘option’ to ‘optional’.

I removed the underscores and rename the helpers (option to optional) in 81628ef

Thanks for the review!

lysnikolaou:
Maybe I'm not getting something, but 'handle_alt_optionals` doesn't sound right. The `Alt` node could have no `Opt`s in it and this function will still be called. Is there some alternate meaning of the word that I'm missing?

pablogsal:
> Maybe I'm not getting something, but 'handle_alt_optionals' doesn't sound right.

Maybe I misunderstood what "option" should I have changed. What about `handle_simple_alt`? Do you have any suggestions for the name?

lysnikolaou:
> What about `handle_simple_alt`?

Sounds better to me.

pablogsal:
@gvanrossum any preference on the name of that helper (the one that does not have `is_loop`)?

lysnikolaou:
> Edit: Originally I used `option` because this path corresponds to things like `A | B` and that is what I was referring to as "option".

Then 'optional` is a good name, but then again it could be something like `A B`  just as well.

pablogsal:
> Then 'optional' is a good name, but then again it could be something like `A B` just as well.

The way I was thinking for that case: the 'option' there is `A B` (1 option) while in `A | B` there are two 'options' (`A` and `B`).

pablogsal:
> So I propose to name this helper function handle_alt_normal.

Done in 3345475!

> We should also see if `is_loop` and `is_gather` are in some way dependent (I suspect `is_gather` requires `is_loop`).

Seems that the only case when `is_gather` is equal to `is_loop` is when both are false as `is_gather=True` is set for gather patterns while both repeat0 and repeat1 generate `is_loop=True`. I think they are independent but I may be missing something.

lysnikolaou:
> Seems that the only case when `is_gather` is equal to `is_loop` is when both are false as `is_gather=True` is set for gather patterns while both repeat0 and repeat1 generate `is_loop=True`. I think they are independent but I may be missing something.

That's right.

PR #156, opened 2020-01-15
Handle missing types in in set_expr_context

### LINKED ISSUE #154: “make simpy_cpython” fails on Ubuntu and with Python built from source
I just found out that `make simpy_cpython` fails on my Ubuntu machine, because the Python version I'm using was build from source `--with-pydebug` and [this](https://github.com/python/cpython/blob/3f12ac18a407983a23d43ae785e805e773571477/Objects/typeobject.c#L3155) assertion fails in `Objects/typeobject.c`.

I think that the file and line that's causing the failure is `Lib/multiprocessing/pool.py` [on line 517](https://github.com/python/cpython/blob/3f12ac18a407983a23d43ae785e805e773571477/Lib/multiprocessing/pool.py#L517).

It seems that this is somehow related to `a = [*b()]` failing to parse at the moment with a `ValueError: field value is required for Starred`, which I cannot figure out the reason for. Any ideas?

pablogsal:
Closes #154

### DIFF
--- a/test/test_ast_generation.py
+++ b/test/test_ast_generation.py
@@ -249,6 +249,7 @@ TEST_CASES = [
         pass; pass
         pass
      '''),
+    ('namedexpr', '(x:=[1, 2, 3])'),
     ('nonlocal', 'nonlocal a, b'),
     ('pass', 'pass'),
     ('pos_args',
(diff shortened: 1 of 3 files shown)


lysnikolaou:
Wouldn't adding `new = expr` in the default case suffice? `Call` objects, for example, will never be targets or star_targets by themselves, so their context doesn't need to be reset.

pablogsal:
> Wouldn't adding `new = expr` in the default case suffice? `Call` objects, for example, will never be targets or star_targets by themselves, so their context doesn't need to be reset.

Right. I think we still need to handle named expressions separately, no?

test/test_ast_generation.py
```diff
@@ -358,6 +358,22 @@ def f(a, b=0, /, c=2):
...
      '''),
+    ('unpacking_binop', '[*([1, 2, 3] + [3, 4, 5])]'),
+    ('unpacking_call', '[*b()]'),
+    ('unpacking_compare', '[*(x < y)]'),
+    ('unpacking_constant', '[*3]'),
+    ('unpacking_dict', '[*{1: 2, 3: 4}]'),
+    ('unpacking_dict_comprehension', '[*{x:y for x,y in z}]'),
+    ('unpacking_ifexpr', '[*([1, 2, 3] if x else y)]'),
+    ('unpacking_list', '[*[1,2,3]]'),
+    ('unpacking_list_compregension', '[*[x for x in y]]'),
```
lysnikolaou: Typo:
```suggestion
    ('unpacking_list_comprehension', '[*[x for x in y]]'),
```

pablogsal:
> I think we still need to handle named expressions separately, no?.

@lysnikolaou After looking at it, the problem was independent: there were no tests with the ':=' in the test suite :fearful:

lysnikolaou:
> > I think we still need to handle named expressions separately, no?.
> 
> @lysnikolaou After looking at it, the problem was independent: there were no tests with the ':=' in the test suite

Oops!

pablogsal:
I have fixed and added a test it in my last ammend (19493a5)

pablogsal:
> LGTM! Thanks a lot for the help with this, again!

Thank you for catching this problem, for the initial investigation and the review! :)

PR #157, opened 2020-01-15
WIP: simpy_cpython runtime logs

lysnikolaou:
I'm opening this WIP PR for the log file with the results of the timings, so that we have a picture until we set up something like `asv`.

### DIFF
--- a/scripts/generate_runtime_logs.py
+++ b/scripts/generate_runtime_logs.py
@@ -1,10 +1,23 @@
+import argparse
 import pathlib
 import subprocess
 
 import git
 
 
+argparser = argparse.ArgumentParser(
+    prog="generate_runtime_logs",
+    description="Helper program to generate runtime logs of running 'make simpy_cpython'",
+)
+argparser.add_argument("-c", "--cpython", help="The directory where CPython resides")
+
+
 def main() -> None:
+    args = argparser.parse_args()
+    cpython_dir = args.cpython or "../cpython"
+    lib_dir = str(pathlib.Path(cpython_dir).joinpath("Lib"))
+    print(lib_dir)
+
     repo = git.Repo(pathlib.Path.cwd())
     with open("simpy_cpython_timing.log", "w") as file:
         for commit in repo.iter_commits("80f8efb..master"):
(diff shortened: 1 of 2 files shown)


lysnikolaou:
Note, that this has `GitPython` as a dependency, which is not listed in `requirements.pip`. 

I can work on doing the necessary changes, in case we want to merge in the script.

gvanrossum:
Thanks! One insight from this is that there's a lot of noise in these numbers. But apart from that, the biggest jump seems to be for 9cda37b0e24a24285cdb5ab6902a1b187391a5a1. I think the biggest change in that commit is that it generates an AST for `primary`, which probably makes the AST much bigger, since so many nodes are created that weren't before, for `Attribute`, `Subscript` and `Call` (which are of course very common). As an experiment I went into that commit and removed the actions for all of `primary`'s alternatives, and most of the speed was given back.

Part of this is AST *construction* (allocating and filling more nodes), but I bet that the bulk of it is *translating* the internal representation of AST nodes to the Python-usable AST nodes (`ast.AST`). This is done by calling `PyAST_mod2obj(res)` in `run_parser()` in `pegen/pegen.c`. This translation process is not free, but we don't care for it from the POV of speeding up the parser, since in *normal* use we skip that translation in favor of compiling to bytecode.

Perhaps we can repeat the experiment with a patch that uses the `mode` flag to suppress the AST translation. (In which case `parse_string()` and `parse_file()` will always return `None`.) Maybe the easiest hack is to remove the type from `start[mod_ty]`.

gvanrossum:
Yeah, a simple test suggests that the cost of `PyAST_mod2obj()` is an order of magnitude more than the cost of constructing the AST nodes for `primary`.

lysnikolaou:
How were the preliminary results mentioned in the README generated? Was it `simpy_cpython` or something else?

gvanrossum:
As the text mentions it was parsing data:xxl.txt.

lysnikolaou:
Yeah, sorry. Didn't make myself clear at all. Were you using some kind of script that would parse data/xxl.txt a few times and would compute the mean or were you parsing it just once?

My point is, if there is such a script, it would be really interesting to run it again and see what the difference in speed is.

gvanrossum:
I don’t recall automating that, sorry.

gvanrossum:
Well, there's `test/timings.py`, although it seems to need a lot of help.
```
python3 -m pegen -c data/simpy.gram --compile
PYTHONPATH=. python3 test/timings.py data/xxl.txt
```

PR #162, opened 2020-01-18
Add --skip-actions flags to generate no (or at least minimal) actions

### LINKED ISSUE #160: add argument to turn off AST generation and measure just parsing
When refactoring PEG grammars there may be rewrites that ease code generation, but hurt the performance of the parser, because PEG is more algorithmic than denotational or functional.

The original `sympy` grammar was performant to the levels of `pgen`, and there are no theoretical reasons why `pegen` can't keep being as performant.

To measure, it's important to have a way to _only_ parse, thus the request for options that make it easy to bypass code generation in unit tests and system tests. After finding rules that are taking more than expected, reordering of clauses and/or lookaheads often return the parser to the previous performance.

(cc @gvanrossum, as discussed on email)

gvanrossum:
Fixes #160 (including the "extra credit" idea of adding another flag to force `mode = 0`).

### DIFF
--- a/pegen/build.py
+++ b/pegen/build.py
@@ -137,7 +137,7 @@ def build_parser_and_generator(
           output when compiling the C extension . Defaults to False.
         keep_asserts_in_extension (bool, optional): Whether to keep the assert statements
           when compiling the extension module. Defaults to True.
-        skip_actions (boo, optional): Whether to pretend no rule has any actions.
+        skip_actions (bool, optional): Whether to pretend no rule has any actions.
     """
     grammar, parser, tokenizer = build_parser(grammar_file, verbose_tokenizer, verbose_parser)
     gen = build_generator(
--- a/pegen/build.py
+++ b/pegen/build.py
@@ -92,13 +92,14 @@ def build_generator(
     compile_extension: bool = False,
     verbose_c_extension: bool = False,
     keep_asserts_in_extension: bool = True,
+    skip_actions: bool = False,
 ) -> ParserGenerator:
     with open(output_file, "w") as file:
         gen: ParserGenerator
         if output_file.endswith(".c"):
-            gen = CParserGenerator(grammar, file)
+            gen = CParserGenerator(grammar, file, skip_actions=skip_actions)
         elif output_file.endswith(".py"):
-            gen = PythonParserGenerator(grammar, file)
+            gen = PythonParserGenerator(grammar, file)  # TODO: skip_actions
         else:
             raise Exception("Your output file must either be a .c or .py file")
         gen.generate(grammar_file)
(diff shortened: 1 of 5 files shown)


gvanrossum:
I hope someone can review this. This does two separate things:

1. Add a `--skip-actions` flag to the generator that replaces all actions (even default ones) with `res = CONSTRUCTOR(p);`, which minimizes the time spent in actions. I also added a little cache to itself so it always returns the same node, minimizing allocations. (This flag currently does nothing when generating Python code.)

2. Change the argument handling for `parse_string()` and `parse_file()` in the generated module so that you can pass a value for `mode`. In particular, `mode=0` suppresses the call to `PyAST_mod2obj()`, so that we can measure the time to construct the internal AST without converting it to the Python-consumable `ast.AST` object type.

pegen/build.py
```diff
@@ -135,6 +137,7 @@ def build_parser_and_generator(
           output when compiling the C extension . Defaults to False.
         keep_asserts_in_extension (bool, optional): Whether to keep the assert statements
           when compiling the extension module. Defaults to True.
+        skip_actions (boo, optional): Whether to pretend no rule has any actions.
```
lysnikolaou:
```suggestion
        skip_actions (bool, optional): Whether to pretend no rule has any actions.
```

pegen/c_generator.py
```diff
@@ -435,14 +453,19 @@ def emit_default_action(self, is_gather: bool, names: List[str], node: Alt) -> N
                 )
             self.print(f"res = {names[0]};")
 
+    def emit_no_action(self) -> None:
```
lysnikolaou: Maybe `emit_dummy_action` would be better?

pegen/c_generator.py
```diff
@@ -29,28 +29,36 @@
 """
 EXTENSION_SUFFIX = """
 static PyObject *
-parse_file(PyObject *self, PyObject *args)
+parse_file(PyObject *self, PyObject *args, PyObject *kwds)
 {
+    static const char *keywords[] = {"file", "mode", NULL};
```
lysnikolaou: What about naming that second keyword argument something other than `mode`? I know, that's what it's called in `pegen.c`, but in the public functions we could name it something like `public_ast` and have it be a `bool`, rather than an `int`.

pegen/pegen.c
```diff
@@ -135,12 +135,17 @@ update_memo(Parser *p, int mark, int type, void *node)
...
+
+    if (cache != NULL)
+        return cache;
 
     PyObject *id = _create_dummy_identifier(p);
     if (!id) {
         return NULL;
     }
-    return Name(id, Load, 1, 0, 1, 0,p->arena);
+    cache = Name(id, Load, 1, 0, 1, 0,p->arena);
```
lysnikolaou:
```suggestion
    cache = Name(id, Load, 1, 0, 1, 0, p->arena);
```

pegen/c_generator.py
gvanrossum: I'd rather keep `mode`, since I intend to expand this to `mode=2` which would compile the code.

PR #164, opened 2020-01-18
By default, compile to bytecode rather than to AST

gvanrossum:
Use mode=1 to return an AST, or mode=0 to return None.

Note that mode=1 is actually the slowest, since it has to convert the
internal AST to an ast.AST tree. mode=0 is the fastest (it creates an
internal AST and then throws it away).  mode=2 is the most realistic,
since this is what import does when loading a module for which no .pyc
file exists.

~This reveals that compiling data/xxl.txt this way is actually faster
than the stdlib compile() function!~
```
make time_stdlib: 1.36 real, 1.09 user, 0.26 sys (median of three)
make time:        0.90 real, 0.68 user, 0.20 sys (median of three). # Never mind, see next comment
```
Also, a run of `make simpy_cpython` now says:
```
Checked 1,641 files, 749,570 lines, 27,622,497 bytes in 6.700 seconds.
That's 111,883 lines/sec, or 4,123,014 bytes/sec.
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -27,16 +27,39 @@ regen-metaparser: pegen/metagrammar.gram pegen/*.py
 # this has different names in different systems so we are abusing the implicit dependency on
 # parse.c by the use of --compile-extension.
 
-test: pegen/parse.c
-	$(PYTHON) -c "from pegen import parse; t = parse.parse_file('$(TESTFILE)'); exec(compile(t, '', 'exec'))"
+.PHONY: test
+
+test: run
+
+run: pegen/parse.c
+	$(PYTHON) -c "from pegen import parse; t = parse.parse_file('$(TESTFILE)'); exec(t)"
 
 compile: pegen/parse.c
-	$(PYTHON) -c "from pegen import parse; t = parse.parse_file('$(TESTFILE)')"
+	$(PYTHON) -c "from pegen import parse; t = parse.parse_file('$(TESTFILE)', mode=2)"
 
-time: pegen/parse.c
-	/usr/bin/time -l $(PYTHON) -c "from pegen import parse; parse.parse_file('$(TIMEFILE)')"
+parse: pegen/parse.c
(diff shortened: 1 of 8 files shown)


gvanrossum:
Note that this depends on #162.

gvanrossum:
Whoops, the comparison to stdlib was unfair, because it used a much simplified grammar (`data/cprog.gram`) instead of the full grammar (`data/simpy.gram`). With the full grammar it's slower:
```
make time_stdlib: 1.36 real, 1.09 user, 0.26 sys (median of three)
make_time:        2.55 real, 2.03 user, 0.52 sys (median of three)
```
In adition, almost twice the amount of memory is used: max RSS is 1575 MiB, while stdlib max RSS is 876 MiB.

gvanrossum:
Also, the segfault in `test/test_c_parser.py` is back. :-(

pablogsal:
> Also, the segfault in `test/test_c_parser.py` is back. :-(

I tried to reproduce locally but as of now, I failed. Can anyone reproduce the segfault on their side?

gvanrossum:
I get the segfault on my Mac every time.

gvanrossum:
Also, since this depends on #162, can you review that?

gvanrossum:
I think I understand the segfaults. It's because the default mode switched up to 2. But the tests in `test/test_c_parser.py` define their own grammar and actions that are supposedly constructing correct AST trees. I guess they don't, at least no well enough to be compiled to bytecode. I could fix it by adding explicit `mode=1` to the `extension.parse_*()` calls, but I think it's better to put something in the header section of the grammar to request compilation, so I'll do that now.

gvanrossum:
(To clarify, this changes the default mode to 1 for ad-hoc grammars like the ones in that test, but still sets the default to 2 for `test/sympy.gram`, by putting `@bytecode True` in there.)

Issue #163, opened 2020-01-18
Strings in AST should not include quotes


gvanrossum:
I just discovered that `make test` prints the output as
```
"Hello ""world"
```
Using bisection I found that it (correctly) printed
```
Hello world
```
before 3be531c (#133). That PR refactored string parsing thoroughly. I presume something went wrong. @pablogsal Can you look into this?

pablogsal:
> Can you look into this?

I did a preliminary investigation and I found that this is something that happens only outside `simpy.gram`. The reason is that the string handling functions for "STRINGS+" need the quotes and the prefixes intact to correctly produce the final string. In simpy.gram this is fine because all strings go trough these pipelines and the final quoting is correct but any grammar that is using `STRING` directly will receive the unrefined string.

Note that even before  3be531c the behaviour of `string_token` is incorrect: If you change `data/cprog.txt` to have prefixes:

```diff
--- a/data/cprog.txt
+++ b/data/cprog.txt
@@ -1,5 +1,5 @@
 if 1:
-    print("Hello " + "world")
+    print(u"Hello " + "world")
     if 0:
```

then `make test` prints the output as (note the initial quote):

```
"Hello world
```

Here is the general problem:

* `STRING` produces code that calls `string_token()` and that produces an ast constant (which is valid by itself) and that *should* not have quotes or prefixes. Currently this nodes contain the RAW string (needed for the second phase) and before commit 3be531c  they contained a string with stripped quotes (but not always correct as the previous example shows).

* When we parse `STRINGS+` we receive multiple constant nodes result of calling  `string_token()` 
 over them but we need the quotes and the prefixes intact to properly produce a final string: a new constant node with the strings joined (this is if we want to reuse the previous machinery as much as possible).

A solution that I can think of initially is somehow producing fully-correct AST nodes in `string_token()` and from them recostructing the original raw string before passing that down to the functions that handle `STRINGS+` (`concatenate_strings() and friends). Care must be taken into consideration as these boxing-unboxing operations can impact performance for no gain (as we are doing operations that need to be undone in later steps).

@gvanrossum @lysnikolaou What do you think the best approach here should be?

gvanrossum:
To summarize, the issue seems to be that `concatenate_strings()` takes a list of `STRING` nodes *with* quotes and returns a single node *without* quotes. And `make test` doesn't use `data/simpy.gram`, it uses `data/cprog.gram`, which doesn't call `concatenate_strings()`.

[...]

PR #165, opened 2020-01-19
Refactor and simplify string token handling and parse them correctly in all grammars

pablogsal:
Closes: #163 

Apart from simplifying things a bit, this PR makes sure that `string_token` *always* produces valid AST string nodes (`Constant`) with the appropriate string types. Then, in `concatenate_strings` we just add them together (if the types are correct) but we don't do any more parsing, which arguably is a better split of responsibilities between the functions.

### DIFF
--- a/pegen/pegen.c
+++ b/pegen/pegen.c
@@ -381,26 +381,76 @@ number_token(Parser *p)
     return Constant(c, NULL, t->lineno, t->col_offset, t->end_lineno, t->end_col_offset, p->arena);
 }
 
+static int
+parsestr(Parser* p, const char* s, int *bytesmode, int *rawmode,
+         PyObject **result, const char **fstr, Py_ssize_t *fstrlen);
+
 expr_ty
 string_token(Parser *p)
 {
     Token *t = expect_token(p, STRING);
-    if (t == NULL)
-        return NULL;
-    char *s = NULL;
-    Py_ssize_t len = 0;
-    if (PyBytes_AsStringAndSize(t->bytes, &s, &len) < 0)
-        return NULL;
-    // TODO: Creating an AST constant here and joining the strings afterwards is
-    // inefficient. We should just move around the char*.
-    PyObject *c = PyUnicode_FromStringAndSize(s, len);
-    if (!c)
-        return NULL;
-    if (PyArena_AddPyObject(p->arena, c) < 0) {
-        Py_DECREF(c);
+    PyObject *u_kind = NULL;
+
+    if (t == NULL) {
         return NULL;
     }
-    return Constant(c, NULL, t->lineno, t->col_offset, t->end_lineno, t->end_col_offset, p->arena);
+
+    char *the_str = PyBytes_AsString(t->bytes) ;
+    if (the_str == NULL) {
+        return NULL;
+    }
+
+    PyObject *s;
+    PyObject *final_str = NULL;
+    int this_rawmode = 0;
+    int bytesmode = 0;
+    const char *fstr;
+    Py_ssize_t fstrlen = -1;  /* Silence a compiler warning. */
+    if (parsestr(p, the_str, &bytesmode, &this_rawmode, &s,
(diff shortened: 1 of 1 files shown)


pablogsal:
With this PR:

```
venv ❯ make test         
python -c "from pegen import parse; t = parse.parse_file('data/cprog.txt'); exec(compile(t, '', 'exec'))"
Hello world
```

pegen/pegen.c
```diff
@@ -381,26 +381,76 @@ number_token(Parser *p)
...
+
+    if (t == NULL) {
         return NULL;
-    // TODO: Creating an AST constant here and joining the strings afterwards is
-    // inefficient. We should just move around the char*.
-    PyObject *c = PyUnicode_FromStringAndSize(s, len);
-    if (!c)
+    }
+
+    char *the_str = PyBytes_AsString(t->bytes) ;
```
gvanrossum:
```suggestion
    char *the_str = PyBytes_AsString(t->bytes);
```

pegen/pegen.c
```diff
@@ -381,26 +381,76 @@ number_token(Parser *p)
...
+    assert(s != NULL && fstr == NULL);
+    assert(bytesmode ? PyBytes_CheckExact(s) : PyUnicode_CheckExact(s));
+    final_str = s;
+
+    if (PyArena_AddPyObject(p->arena, final_str) < 0) {
+        goto error;
+    }
+
+    if (!bytesmode && kind_unicode) {
+        //TODO: Intern this string when we decide how we will
```
gvanrossum: Do you intentionally leave out the space between `//` and `TODO`?

pegen/pegen.c
```diff
@@ -381,26 +381,76 @@ number_token(Parser *p)
...
+
+    char *the_str = PyBytes_AsString(t->bytes) ;
+    if (the_str == NULL) {
         return NULL;
-    if (PyArena_AddPyObject(p->arena, c) < 0) {
-        Py_DECREF(c);
+    }
+
+    PyObject *s;
+    PyObject *final_str = NULL;
```
gvanrossum: Why not move this declaration to the first use at line 434?

pegen/pegen.c
```diff
@@ -381,26 +381,76 @@ number_token(Parser *p)
...
+    if (parsestr(p, the_str, &bytesmode, &this_rawmode, &s,
+                 &fstr, &fstrlen) != 0) {
         return NULL;
     }
-    return Constant(c, NULL, t->lineno, t->col_offset, t->end_lineno, t->end_col_offset, p->arena);
+
+    /* Check if it has a 'u' prefix */
+    int kind_unicode = 0;
+    if (the_str[0] == 'u') {
+        kind_unicode = 1;
```
gvanrossum: Maybe `assert(!bytesmode);` here? They're mutually exclusive.

pegen/pegen.c
```diff
@@ -381,26 +381,76 @@ number_token(Parser *p)
...
 expr_ty
 string_token(Parser *p)
 {
     Token *t = expect_token(p, STRING);
-    if (t == NULL)
-        return NULL;
-    char *s = NULL;
-    Py_ssize_t len = 0;
-    if (PyBytes_AsStringAndSize(t->bytes, &s, &len) < 0)
+    PyObject *u_kind = NULL;
```
gvanrossum: You can move this to line 439. (I like declaring variables near the first point they're needed.)

pegen/pegen.c
pablogsal: Nop, there should be a space there.

pablogsal:
> Is there even a point in preserving the 'u' flag, given that we're only Parsing Python 3 and it's redundant?

Functionally not much, but helps with testing (more opportunities to test prefixes) and to mirror what the current AST in CPython produces (the `kind` field in string nodes). We could remove it in the future to simplify the code, indeed.

PR #167, opened 2020-01-19
Set tok->filename to '<string>' in run_parser_from_string()

### LINKED ISSUE #166: parse_string segfaults when using mode=2
Due to `tok->filename` being NULL, when `parse_string(..., mode=2)` is called, [Python/compile.c:336](https://github.com/python/cpython/blob/558f07891170fe5173f277d3749e92d844de0a27/Python/compile.c#L336) segfaults. What's the best course of action here?

gvanrossum:
Fixes #166

### DIFF
--- a/pegen/pegen.c
+++ b/pegen/pegen.c
@@ -532,7 +532,16 @@ run_parser_from_string(const char* str, void *(start_rule_func)(Parser *), int m
     if (tok == NULL)
         return NULL;
 
-    PyObject* result = run_parser(tok, start_rule_func, mode);
+    // From here on we need to clean up even if there's an error
+    PyObject *result = NULL;
+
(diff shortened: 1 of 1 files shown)


gvanrossum:
If this passes tests and I'm not responding right away please land it for me.

lysnikolaou:
I was just trying to fix this and I was also getting this segfault when parsing `del a()` with mode=0, which seems to happen due to `raise_syntax_error` relying on tok->filename being `NULL`. I can take it from here, if you don't have time.

gvanrossum:
Oh, I see. We shouldn't set `tok->filename`, but we should set the `filename` passed into `PyAST_CompileObject()` in `run_parser()`. Why don't you tackle that, I'll just close my PR.

PR #168, opened 2020-01-19
Set filename to '<string>' in run_parser in case tok->filename is NULL

### LINKED ISSUE #166: parse_string segfaults when using mode=2

lysnikolaou:
Fixes #166.

### DIFF
--- a/pegen/pegen.c
+++ b/pegen/pegen.c
@@ -469,7 +469,12 @@ run_parser(struct tok_state* tok, void *(start_rule_func)(Parser *), int mode)
     }
 
     if (mode == 2) {
-        result = (PyObject *)PyAST_CompileObject(res, tok->filename, NULL, -1, p->arena);
+        PyObject *filename = (tok->filename)
+                             ? tok->filename
+                             : PyUnicode_FromString("<string>");
+        if (!filename)
+            goto exit;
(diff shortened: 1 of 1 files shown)


lysnikolaou:
> This will leak the object created.

Is it okay if I add a `Py_XDECREF` after `PyAST_CompileObject` has been called?

gvanrossum:
> Is it gonna be okay if I add a Py_XDECREF after PyAST_CompileObject has been called?

Sure, but only if PyUnicode_FromString() was called -- otherwise you'd be stealing a ref from `tok`.

pegen/pegen.c
```diff
@@ -469,7 +469,14 @@ run_parser(struct tok_state* tok, void *(start_rule_func)(Parser *), int mode)
     }
 
     if (mode == 2) {
-        result = (PyObject *)PyAST_CompileObject(res, tok->filename, NULL, -1, p->arena);
+        PyObject *filename = (tok->filename)
```
gvanrossum:
```suggestion
        PyObject *filename = tok->filename
```

gvanrossum:
Thanks!

PR #169, opened 2020-01-19
Check for reserved keywords in name_token

### LINKED ISSUE #152: Handle reserved words like the old parser
In the old parser, reserved words (e.g. `if`) are *always* reserved. We should mimic this behavior for most keywords, otherwise people will start abusing the feature. But there should be a way to turn it off for specific situations, to support e.g. `async def` and `await` (and in the future to support other "compound keywords", like `match of`, if we want to go there).

lysnikolaou:
Also return Constant nodes for True, False and None.

Closes #152.

### DIFF
--- a/pegen/pegen.c
+++ b/pegen/pegen.c
@@ -304,6 +304,49 @@ endmarker_token(Parser *p)
     return expect_token(p, ENDMARKER);
 }
 
+static const char *reserved_keywords[] = {
+    "and",
+    "as",
+    "assert",
+    "async",
+    "await",
+    "break",
+    "class",
+    "continue",
+    "def",
+    "del",
+    "elif",
+    "else",
+    "except",
+    "finally",
+    "for",
+    "from",
+    "global",
+    "if",
+    "import",
+    "in",
+    "is",
+    "lambda",
+    "nonlocal",
+    "not",
+    "or",
+    "pass",
+    "raise",
+    "return",
+    "try",
+    "while",
+    "with",
+    "yield"
+};
+static int // bool
+_is_reserved_keyword(char *s) {
+    for (int i = 0; i < 32; i++) {
+        if (s[0] == reserved_keywords[i][0] && !strcmp(s, reserved_keywords[i]))
+            return 1;
+    }
+    return 0;
+}
+
 expr_ty
 name_token(Parser *p)
 {
@@ -313,6 +356,17 @@ name_token(Parser *p)
     char *s;
     Py_ssize_t n;
     if (PyBytes_AsStringAndSize(t->bytes, &s, &n) < 0)
+        return NULL;
+    if (!strcmp(s, "None"))
+        return Constant(Py_None, NULL, t->lineno, t->col_offset,
+                        t->end_lineno, t->end_col_offset, p->arena);
+    if (!strcmp(s, "True"))
+        return Constant(Py_True, NULL, t->lineno, t->col_offset,
+                        t->end_lineno, t->end_col_offset, p->arena);
+    if (!strcmp(s, "False"))
+        return Constant(Py_False, NULL, t->lineno, t->col_offset,
+                        t->end_lineno, t->end_col_offset, p->arena);
+    if (_is_reserved_keyword(s))
         return NULL;
     PyObject *id = PyUnicode_DecodeUTF8(s, n, NULL);
     if (id == NULL)
--- a/pegen/pegen.c
+++ b/pegen/pegen.c
@@ -341,7 +341,7 @@ static const char *reserved_keywords[] = {
 static int // bool
 _is_reserved_keyword(char *s) {
     for (int i = 0; i < 32; i++) {
-        if (s[0] == reserved_keywords[i][0] && !strcmp(s, reserved_keywords[i]))
+        if (!strcmp(s, reserved_keywords[i]))
             return 1;
     }
     return 0;
--- a/test/test_ast_generation.py
+++ b/test/test_ast_generation.py
@@ -252,7 +252,10 @@ TEST_CASES = [
         pass; pass
         pass
      '''),
-    ('namedexpr', '(x:=[1, 2, 3])'),
+    ('namedexpr', '(x := [1, 2, 3])'),
+    ('namedexpr_false, (x := False)'),
+    ('namedexpr_none', '(x := None)'),
+    ('namedexpr_true', '(x := True)'),
     ('nonlocal', 'nonlocal a, b'),
     ('pass', 'pass'),
     ('pos_args',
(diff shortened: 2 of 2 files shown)


pegen/pegen.c
```diff
@@ -304,6 +304,49 @@ endmarker_token(Parser *p)
...
+    "return",
+    "try",
+    "while",
+    "with",
+    "yield"
+};
+static int // bool
+_is_reserved_keyword(char *s) {
+    for (int i = 0; i < 32; i++) {
+        if (s[0] == reserved_keywords[i][0] && !strcmp(s, reserved_keywords[i]))
```
pablogsal: Why `s[0] == reserved_keywords[i][0] `? `strcmp` is normally very optimized and has early exits. For example:

https://github.com/bminor/glibc/blob/5f72f9800b250410cad3abfeeb09469ef12b2438/string/strcmp.c#L30-L46

test/test_ast_generation.py
```diff
@@ -19,6 +19,9 @@
     ('annotation_with_parens', '(parens): int'),
     ('assert', 'assert a'),
     ('assert_message', 'assert a, b'),
+    ('assignment_false', 'a = False'),
```
pablogsal: I would recommend adding some tests with `:=`

pegen/pegen.c
lysnikolaou: It's what [Parser/parse.c](https://github.com/python/cpython/blob/558f07891170fe5173f277d3749e92d844de0a27/Parser/parser.c#L151) does. But you are right! I'll remove it.

pablogsal: Now I am curious about performance benchmarks about that....maybe I am missing something. But I would advocate starting with cleaner code before we start optimizing everything :)

test/test_ast_generation.py
```diff
@@ -249,7 +252,10 @@ def f(a, b=2, /, c=3, *e, f, **g):
         pass; pass
         pass
      '''),
-    ('namedexpr', '(x:=[1, 2, 3])'),
+    ('namedexpr', '(x := [1, 2, 3])'),
+    ('namedexpr_false, (x := False)'),
```
pablogsal:
```suggestion
    ('namedexpr_false', '(x := False)'),
```

pablogsal:
```suggestion
    ('namedexpr_false', '(x := False)'),
```

pablogsal:
```suggestion
    ('namedexpr_false', '(x := False)'),
```

pegen/pegen.c
```diff
@@ -341,7 +341,7 @@ static const char *reserved_keywords[] = {
 static int // bool
 _is_reserved_keyword(char *s) {
     for (int i = 0; i < 32; i++) {
-        if (s[0] == reserved_keywords[i][0] && !strcmp(s, reserved_keywords[i]))
```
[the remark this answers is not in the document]
pablogsal: I would say yes, but maybe let's group all PEP7 fixes in a new PR

pegen/pegen.c
```diff
@@ -304,6 +304,49 @@ endmarker_token(Parser *p)
     return expect_token(p, ENDMARKER);
 }
 
+static const char *reserved_keywords[] = {
+    "and",
```
pablogsal: Idea: Maybe we should have a way to automatically generate this list from the grammar. Maybe we could make this list available from the generated code...

gvanrossum: Honestly I'm not sure if that's worth the effort. Sure, `pgen` generates the list of keywords automatically from the grammar. But that harks back to a time when adding a new keyword was no big deal. The last time we added new keywords it was a big mess (`async`/`await`) and those still aren't proper keywords (due to the `feature_version` flag on `ast.parse()`).

pegen/pegen.c
```diff
@@ -304,6 +304,49 @@ endmarker_token(Parser *p)
...
+    "return",
+    "try",
+    "while",
+    "with",
+    "yield"
+};
+static int // bool
+_is_reserved_keyword(char *s) {
+    for (int i = 0; i < 32; i++) {
+        if (!strcmp(s, reserved_keywords[i]))
```
gvanrossum: I find the spelling `!strcmp(a, b)` confusing, since `!` makes my brain say *not*. Even though it's more typing, can you change to `strcmp(a, b) == 0`? I'm sure the C compiler optimizes it the right way.

pegen/pegen.c
```diff
@@ -314,6 +357,17 @@ name_token(Parser *p)
...
+        return Constant(Py_True, NULL, t->lineno, t->col_offset,
+                        t->end_lineno, t->end_col_offset, p->arena);
+    if (!strcmp(s, "False"))
+        return Constant(Py_False, NULL, t->lineno, t->col_offset,
+                        t->end_lineno, t->end_col_offset, p->arena);
+    if (_is_reserved_keyword(s))
+        return NULL;
     PyObject *id = PyUnicode_DecodeUTF8(s, n, NULL);
     if (id == NULL)
         return NULL;
```
gvanrossum: I think the `// TODO` below on L378 is obsolete?

pegen/pegen.c
```diff
@@ -304,6 +304,49 @@ endmarker_token(Parser *p)
...
+    "nonlocal",
+    "not",
+    "or",
+    "pass",
+    "raise",
+    "return",
+    "try",
+    "while",
+    "with",
+    "yield"
```
gvanrossum: Trailing comma.

pegen/pegen.c
```diff
@@ -304,6 +304,49 @@ endmarker_token(Parser *p)
...
+};
+static int // bool
+_is_reserved_keyword(char *s) {
+    for (int i = 0; i < 32; i++) {
```
gvanrossum: While I am okay with updating the array of reserved words manually, I am not okay with having the number of reserved words as a literal in the code. Maybe you can append a `NULL` to the array as a sentinel.

lysnikolaou:
> I don't think it's too soon for optimization. This seems to cause a 5% slowdown.
> 
> One thing that worries me in particular: the cache is ineffective for checking keywords, since it only stores the type returned by the tokenizer, not the info gleaned from checking for keywords. So that means if we backtrack a few times over a `NAME` token we do all the keyword checks. (And if it is one of `None`, `True`, `False`, we also create a new `Constant` object each time!)
> 
> IIRC pgen solves this differently -- it assigns a different token type to each keyword (though admittedly I've forgotten the details :-( ). If we did a similar thing, we could check for keywords once, in `fill_token()`. At this point it may also be necessary to generate the array from `c_generate.py`, so that it can know the assigned token type for each keyword, and we could replace all `expect_keyword()` calls with `expect_token()`. (N/T/F would be handled by semantic actions for these as alternatives for `atom`.

I really like this approach. Working on it!

lysnikolaou:
I'll close this PR and start a new one, when I'm ready.

PR #170, opened 2020-01-19
Remove dead link from README

lysnikolaou:


lysnikolaou:
I'll go ahead and land this! In case you have objections, we can always revert it.

PR #174, opened 2020-01-21
Generate different token type for each keyword

### LINKED ISSUE #152: Handle reserved words like the old parser

lysnikolaou:
This way we can correctly handle parsing keywords. This includes:
    * Throwing an exception if a keyword is used as a NAME
    * Generating correct nodes for True, False, None

Closes #152.

PS: Mypy is not happy for some reason. Any suggestions on what to do to get rid of the errors?

### DIFF
--- a/pegen/c_generator.py
+++ b/pegen/c_generator.py
@@ -99,12 +99,9 @@ class CCallMakerVisitor(GrammarVisitor):
         if name in ("NAME", "NUMBER", "STRING"):
             name = name.lower()
             return f"{name}_var", f"{name}_token(p)"
-        if name in ("NEWLINE", "DEDENT", "INDENT", "ENDMARKER"):
+        if name in ("NEWLINE", "DEDENT", "INDENT", "ENDMARKER", "ASYNC", "AWAIT"):
             name = name.lower()
             return f"{name}_var", f"{name}_token(p)"
-        if name in ("ASYNC", "AWAIT"):
-            name = name.lower()
-            return self.keyword_helper(name)
         return f"{name}_var", f"{name}_rule(p)"
 
     def visit_StringLeaf(self, node: StringLeaf) -> Tuple[str, str]:
@@ -191,7 +188,7 @@ class CParserGenerator(ParserGenerator, GrammarVisitor):
         skip_actions: bool = False,
     ):
         super().__init__(grammar, file)
-        self.callmakervisitor = CCallMakerVisitor(self)
+        self.callmakervisitor: CCallMakerVisitor = CCallMakerVisitor(self)
         self._varname_counter = 0
         self.debug = debug
         self.skip_actions = skip_actions
--- a/pegen/c_generator.py
+++ b/pegen/c_generator.py
@@ -87,21 +87,30 @@ class CCallMakerVisitor(GrammarVisitor):
     def __init__(self, parser_generator: ParserGenerator):
         self.gen = parser_generator
         self.cache: Dict[Any, Any] = {}
+        self.keyword_cache: Dict[str, int] = {}
+
+    def keyword_helper(self, keyword: str) -> Tuple[str, str]:
+        if keyword not in self.keyword_cache:
+            self.keyword_cache[keyword] = self.gen.keyword_type()
+        return "keyword", f"expect_token(p, {self.keyword_cache[keyword]})"
 
     def visit_NameLeaf(self, node: NameLeaf) -> Tuple[str, str]:
         name = node.value
         if name in ("NAME", "NUMBER", "STRING"):
             name = name.lower()
             return f"{name}_var", f"{name}_token(p)"
-        if name in ("NEWLINE", "DEDENT", "INDENT", "ENDMARKER", "ASYNC", "AWAIT"):
+        if name in ("NEWLINE", "DEDENT", "INDENT", "ENDMARKER"):
             name = name.lower()
             return f"{name}_var", f"{name}_token(p)"
+        if name in ("ASYNC", "AWAIT"):
+            name = name.lower()
+            return self.keyword_helper(name)
         return f"{name}_var", f"{name}_rule(p)"
 
     def visit_StringLeaf(self, node: StringLeaf) -> Tuple[str, str]:
         val = ast.literal_eval(node.value)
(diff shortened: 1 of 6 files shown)


gvanrossum:
To fix the mypy error, add
```
    callmakervisitor: CCallMakerVisitor
```
at the class level to `CParserGenerator`. Otherwise mypy assumes it has the type defined in `parser_generator.py`, i.e. `GrammarVisitor` (which doesn't have the attribute).

pegen/c_generator.py
```diff
@@ -87,21 +87,30 @@ class CCallMakerVisitor(GrammarVisitor):
...
         if name in ("NAME", "NUMBER", "STRING"):
             name = name.lower()
             return f"{name}_var", f"{name}_token(p)"
-        if name in ("NEWLINE", "DEDENT", "INDENT", "ENDMARKER", "ASYNC", "AWAIT"):
+        if name in ("NEWLINE", "DEDENT", "INDENT", "ENDMARKER"):
             name = name.lower()
             return f"{name}_var", f"{name}_token(p)"
+        if name in ("ASYNC", "AWAIT"):
+            name = name.lower()
+            return self.keyword_helper(name)
```
gvanrossum: But `ASYNC` and `AWAIT` are tokens returned by the tokenizer -- they don't need the keyword treatment here. I think you can just keep the old code for these.

pegen/pegen.c
```diff
@@ -153,6 +158,20 @@ CONSTRUCTOR(Parser *p, ...)
...
 }
 
+static int
+_get_keyword_or_name_type(Parser *p, char *name, size_t name_len)
+{
+    for (int i = 0; i < p->n_keywords; i++) {
+        KeywordToken k = (*p->keywords)[i];
+        if (k.str[0] == name[0]
+            && strlen(k.str) == name_len
+            && strncmp(k.str, name, name_len) == 0) {
```
gvanrossum: This sounds convoluted. As Pablo mentioned, `strcmp()` is highly optimized (basically the compiler inlines tricks like these) so you're probably better off just writing only `if (strcmp(k.str, name)) {`. (Unless you've timed it already?)

If you want to try a different kind of optimization, maybe you could generate separate arrays for each keyword length, so e.g. if the name has length 1 you know it's not a keyword, if it has length 2 you only compare it to `if`, `is`, `in`, `as`, `or`, and so on.

pegen/pegen.c
```diff
@@ -182,7 +201,9 @@ fill_token(Parser *p)
     }
 
     Token *t = p->tokens[p->fill];
-    t->type = type;
+    t->type = (type == NAME || type == ASYNC || type == AWAIT)
```
gvanrossum: Don't check for ASYNC/AWAIT here?

pegen/pegen.c
```diff
@@ -296,18 +317,6 @@ get_last_nonnwhitespace_token(Parser *p)
...
-{
-    return expect_token(p, ASYNC);
-}
-
-void *
-await_token(Parser *p)
-{
-    return expect_token(p, AWAIT);
-}
-
```
gvanrossum: Keep these.

pegen/pegen.c
```diff
@@ -517,11 +526,6 @@ run_parser(struct tok_state *tok, void *(start_rule_func)(Parser *), int mode)
     if (!p->arena) {
         goto exit;
     }
-
-    if (fill_token(p) < 0) {
-        goto exit;
-    }
-
```
gvanrossum: Curious why this can be deleted?

lysnikolaou: This actually has to be deleted, since at the time this is called `start_rule` has not been called yet, which means that `init_keywords` hasn't been called as well. Thus calling `fill_token` here would segfault.

pegen/c_generator.py
```diff
@@ -396,6 +410,11 @@ def visit_Rule(self, node: Rule) -> None:
             self._set_up_rule_memoization(node, result_type)
 
         self.print("{")
+        if node.name.startswith("start"):
+            with self.indent():
+                self.print(
+                    f"init_keywords(p, &keywords, {len(self.callmakervisitor.keyword_cache)});"
+                )
```
gvanrossum: Its unfortunate that we have to do this in a specific rule, since in the future we need to be able to use different start rules for the same grammar (e.g. single_input, file_input, eval_input in the pgen Grammar file). But the only way I can think of is to pass the list of keywords to the run_parser* functions. What do you think?

lysnikolaou: Hadn't thought of this! It seems like a better approach for sure. Should I update this PR?

pegen/pegen.c
lysnikolaou: This is not an optimization. This is actually needed, because `name` does not only contain the current token, but everything from the current token until the end of input. This means that `strcmp` would fail, even if the current token is indeed a keyword.

gvanrossum: OK, got it. I think it's worth a comment. (Also, a future optimization would be to have separate lists of keywords by length.)

pegen/c_generator.py
gvanrossum: Go for it.

lysnikolaou:
Thanks for the help!

apalala:
My apologies for having arrived late to this discussion.

Solving keywords versus identifier ambiguity and other complex semantics becomes much easier if there is a way for the semantic action to _fail_ the rule or option. Thus, the rule for `IDENTIFIER` can do:
```python
if is_keyword(name) {
    return SEMANTIC_ERROR  # or whatever
}
```

Trying to solve all complex cases in the grammar or in post-processing leads to convoluted grammars and/or postprocessing.

apalala:
Actually, the above wouldn't work within the `pegen` scheme of semantic actions.

For keywords, TatSu has a special tag for rules that correspond to identifiers, and the parsing engine fails the parse if the parsed identifier is a keyword. Some similar, simple shortcut can be taken in `pegen`.

PR #177, opened 2020-01-22
Teach pegen to compute first sets

pablogsal:


gvanrossum:
How are you planning to use this?

pablogsal:
> How are you planning to use this?

Is a first step towards exploring what was discussed in https://github.com/gvanrossum/pegen/issues/49 regarding [this paper](http://kmizu.github.io/papers/paste513-mizushima.pdf). Ideally, we could also use this information for introducing lookaheads. Mainly I thought is nice to have to also compare with the original grammar

gvanrossum:
Hm, #49 is mostly about something with 'cut' that I fail to understand, and I'm not sure I follow the paper either. But I do have an idea for using FIRST sets to optimize the parser, and this is a first step.

However, an important reason why FIRST set calculation is subtle is that it should properly handle rules with empty productions. I have this example:
```
start: sign thing $
sign: ['-']
thing: NUMBER
```
and your program claims that the FIRST set for `start` is just `{"'-'"}`, whereas it should include `NUMBER` as well. Also, the FIRST set for `sign` should indicate that it may be empty. I think traditionally there's a symbol "epsilon" that's used for this; we could use either an empty string or `None`.

(You could argue that there's something wrong with this grammar, because we don't expect empty productions, but I think it may be better to just deal with those -- it's slightly complicated to prove that a rule *doesn't* have an empty production.)

gvanrossum:
FWIW, parser_generator.py has a function compute_nullables() that can help with this.

apalala:
Please see [this pull request on TatSu](https://github.com/neogeny/TatSu/pull/139) for a full discussion, and a promising implementation.

I agree with @gvanrossum that his simplification of the nullability algorithm is enough to detect and resolve left-rec.

gvanrossum:
@pablogsal Are you waiting for someone or something?

pablogsal:
> @pablogsal Are you waiting for someone or something?

No, I was finishing other stuff. I will land this now.

pegen/first_sets.py
```diff
@@ -0,0 +1,153 @@
...
+
+            # If the set of new terminals can start with the empty string,
+            # it means that the item is completelly nullable and we should
+            # also considering at least the next item in case the current
+            # one fails to parse.
+
+            if "" in new_terminals:
+                continue
+
+            if not isinstance(other.item, (Opt, NegativeLookahead, Repeat0)):
```
apalala: This PR has already been merged, but `Cut` seems to be missing from this list.

pegen/first_sets.py
```diff
@@ -0,0 +1,153 @@
...
+            result |= new_terminals
+            if to_remove:
+                result -= to_remove
+
+            # If the set of new terminals can start with the empty string,
+            # it means that the item is completelly nullable and we should
+            # also considering at least the next item in case the current
+            # one fails to parse.
+
+            if "" in new_terminals:
```
apalala: Perhaps:
```python
if "" not in new_terminals:
    break
```

PR #179, opened 2020-01-24
Generate a 2D array which arranges the keywords by length

### LINKED ISSUE #175: Optimization: Multiple reserved keyword lists by length
For future reference.

As proposed by @gvanrossum in #174.

lysnikolaou:
Closes #175.


lysnikolaou:
> I don't measure a significant perf improvement, unfortunately.

I don't think this would hurt either though.

Issue #180, opened 2020-01-25
F-string support


gvanrossum:
We still don't support f-strings -- an AST is generated with the string constant `"f-strings not supported yet!!"`.

(This is what's left over of #127.)

pablogsal:
@lysnikolaou What do you think if I open a WIP PR to work together on this? With care, we could both push to the PR to iterate. This requires a non-trivial amount of work and I feel I don't have as much time as I wanted so we could finish it together to speed things up

lysnikolaou:
> @lysnikolaou What do you think if I open a WIP PR to work together on this? With care, we could both push to the PR to iterate. This requires a non-trivial amount of work and I feel I don't have as much time as I wanted so we could finish it together to speed things up

Of course! Go ahead and open a PR and let's talk the details over there.

PR #183, opened 2020-01-26
Add f-string support

pablogsal:
Closes: #180 

Things to do:

- [x]  Support parsing single f-strings (no offset info).
- [x] Fix offsets and line numbers for the single f-string case.
- [x] Support concatenation of f-strings.
- [x] Clean code: the f-string parser is prepared to concatenate strings and f-strings together but with the pegen layout, this can be simplified as we need to already produce valid string nodes in `string_token`.
- [ ] Double-check error checking code.
- [x] Add new tests: ideally multiple kinds of expressions so we cover all possibilities for the code that calculates the offsets.


pablogsal:
@lysnikolaou You can push directly to the PR if you push to my branch in my fork:

```
git remote add pablogsal git@github.com:pablogsal/pegen.git
git push pablogsal HEAD:fstrings
```

pegen/pegen.c
```diff
@@ -1690,7 +2595,10 @@ concatenate_strings(Parser *p, asdl_seq *strings)
     PyObject *u_kind = NULL;
     int kind_unicode = 0;
     PyObject *final_str = NULL;
+    return first;
```
pablogsal: This is temporary until the rest of the f-string code is correct. It allows executing the f-string test without worrying of concatenation.

pablogsal:
Ok, executing `python -m pytest test -vvvv -x -k fstring -s` with the PR gives us this:

```
E       assert 'Module(body=...e_ignores=[])' == 'Module(body=...e_ignores=[])'
E         - Module(body=[Expr(value=JoinedStr(values=[FormattedValue(value=BinOp(left=Name(id='val', ctx=Load(), lineno=1, col_offset=1, end_lineno=1, end_col_offset=4), op=Add(), right=Constant(value=2, kind=None, lineno=1, col_offset=5, end_lineno=1, end_col_offset=6), lineno=1, col_offset=1, end_lineno=1, end_col_offset=6), conversion=-1, format_spec=None, lineno=1, col_offset=1, end_lineno=1, end_col_offset=1)], lineno=1, col_offset=0, end_lineno=1, end_col_offset=10), lineno=1, col_offset=0, end_lineno=1, end_col_offset=10)], type_ignores=[])
E         + Module(body=[Expr(value=JoinedStr(values=[FormattedValue(value=BinOp(left=Name(id='val', ctx=Load(), lineno=1, col_offset=3, end_lineno=1, end_col_offset=6), op=Add(), right=Constant(value=2, kind=None, lineno=1, col_offset=7, end_lineno=1, end_col_offset=8), lineno=1, col_offset=3, end_lineno=1, end_col_offset=8), conversion=-1, format_spec=None, lineno=1, col_offset=0, end_lineno=1, end_col_offset=10)], lineno=1, col_offset=0, end_lineno=1, end_col_offset=10), lineno=1, col_offset=0, end_lineno=1, end_col_offset=10)], type_ignores=[])
```

So (without considering string concatenation for now) the only thing it seems that we need to fix is the offsets.

pablogsal:
@lysnikolaou could you take a look at the node locations and offsets? If I am not mistaken, it should be adapting the currently commented `fstring_fix_node_location`.

lysnikolaou:
> @lysnikolaou could you take a look at the node locations and offsets? If I am not mistaken, it should be adapting the currently commented `fstring_fix_node_location`.

On it!

pegen/pegen.c
```diff
@@ -1677,6 +1615,973 @@ parsestr(Parser *p, const char *s, int *bytesmode, int *rawmode, PyObject **resu
...
+    PyMem_Free(p2->tokens);
+    PyMem_Free(p2);
+    if (mod == NULL) {
+        return NULL;
+    }
+    if (asdl_seq_LEN(mod->v.Module.body) != 1) {
+        raise_syntax_error(p, "f-string: invalid expression");
+    }
+    expr_ty expr = asdl_seq_GET(mod->v.Module.body, 0);
+    return expr->v.Subscript.value;
```
lysnikolaou: Why are you using `Subscript` here?

lysnikolaou:
@pablogsal I'm also having some trouble understanding how to get to the children of a `Token`. Is there some hidden structure in `p->tokens` that I fail to understand or do I have to come up with an approach for doing this?

pegen/pegen.c
pablogsal: Because for some reason that I am currently investigating, the first  (and only) expression in the `mod` is enclosed in a `Subscript`:

```
(gdb) p mod
$4 = (mod_ty) 0x555555d6ff78
(gdb) p mod.kind 
$5 = Module_kind
(gdb) p expr
$6 = (expr_ty) 0x555555d6f5d0
(gdb) p expr.kind
$7 = Subscript_kind
```

help to understand what is going on here would be appreciated ;)

pablogsal:
> @pablogsal I'm also having some trouble understanding how to get to the children of a `Token`. Is there some hidden structure in `p->tokens` that I fail to understand or do I have to come up with an approach for doing this?

Before the parser was shifting the CST nodes to correct the already-parsed offsets and line numbers to account for the location of the fstring itself (done with `fstring_fix_node_location`). Now, we don't have such nodes but we only have: the token information for the string itself (which gives us the position of the fstring) and the `expr_ty` resulting of parsing the f-string. The idea is modifying `fstring_fix_node_location` (or create a new analogous function) to modify recursively the offsets and line numbers of the `expr_ty` with the information that we have available (that comes from the token itself). The token probably needs to be threaded down to that function.

pegen/pegen.c
pablogsal: Oh, is because `asdl_seq_GET(mod->v.Module.body, 0)` should be `stmt_ty` not `expr_ty`. Pushing a fix

lysnikolaou:
@pablogsal I just pushed the needed changes, in order for the expr locations to be correct. After some minor testing locally, this appears to be working properly.

Also, I messed up the git history because of my merges. Could I do something to clean this up?

lysnikolaou:
Idea: How would you like moving all the string parsing code to a new file?

pablogsal:
> Idea: How would you like moving all the string parsing code to a new file?

I agree, but I would suggest to so it in the cleanup phase of the PR because the string stuff needs some parser functions so we need some forward declarations and headers prototypes to avoid circular dependencies.

pablogsal:
@lysnikolaou Something is odd with the `col_offset ` of calls. Consider for example:

```
"f'{val+1*34+f(x)
```

We get this error (the `col_offset` of `x` inside the call is incorrect):

```
E         - Module(body=[Expr(value=JoinedStr(values=[FormattedValue(value=BinOp(left=BinOp(left=Name(id='val', ctx=Load(), lineno=1, col_offset=3, end_lineno=1, end_col_offset=6), op=Add(), right=BinOp(left=Constant(value=1, kind=None, lineno=1, col_offset=7, end_lineno=1, end_col_offset=8), op=Mult(), right=Constant(value=34, kind=None, lineno=1, col_offset=9, end_lineno=1, end_col_offset=11), lineno=1, col_offset=7, end_lineno=1, end_col_offset=11), lineno=1, col_offset=3, end_lineno=1, end_col_offset=11), op=Add(), right=Call(func=Name(id='f', ctx=Load(), lineno=1, col_offset=12, end_lineno=1, end_col_offset=13), args=[Name(id='x', ctx=Load(), lineno=1, col_offset=12, end_lineno=1, end_col_offset=13)], keywords=[], lineno=1, col_offset=12, end_lineno=1, end_col_offset=16), lineno=1, col_offset=3, end_lineno=1, end_col_offset=16), conversion=-1, format_spec=None, lineno=1, col_offset=0, end_lineno=1, end_col_offset=18)], lineno=1, col_offset=0, end_lineno=1, end_col_offset=18), lineno=1, col_offset=0, end_lineno=1, end_col_offset=18)], type_ignores=[])
E         + Module(body=[Expr(value=JoinedStr(values=[FormattedValue(value=BinOp(left=BinOp(left=Name(id='val', ctx=Load(), lineno=1, col_offset=3, end_lineno=1, end_col_offset=6), op=Add(), right=BinOp(left=Constant(value=1, kind=None, lineno=1, col_offset=7, end_lineno=1, end_col_offset=8), op=Mult(), right=Constant(value=34, kind=None, lineno=1, col_offset=9, end_lineno=1, end_col_offset=11), lineno=1, col_offset=7, end_lineno=1, end_col_offset=11), lineno=1, col_offset=3, end_lineno=1, end_col_offset=11), op=Add(), right=Call(func=Name(id='f', ctx=Load(), lineno=1, col_offset=12, end_lineno=1, end_col_offset=13), args=[Name(id='x', ctx=Load(), lineno=1, col_offset=14, end_lineno=1, end_col_offset=15)], keywords=[], lineno=1, col_offset=12, end_lineno=1, end_col_offset=16), lineno=1, col_offset=3, end_lineno=1, end_col_offset=16), conversion=-1, format_spec=None, lineno=1, col_offset=0, end_lineno=1, end_col_offset=18)], lineno=1, col_offset=0, end_lineno=1, end_col_offset=18), lineno=1, col_offset=0, end_lineno=1, end_col_offset=18)], type_ignores=[])
```

lysnikolaou:
> @lysnikolaou Something is odd with the `col_offset ` of calls. Consider for example:
> 
> ```
> f'{val+1*34+f(x)}'
> ```
> 
> We get this error (the `col_offset` of `x` inside the call is incorrect):
> 
> ```
> E         - Module(body=[Expr(value=JoinedStr(values=[FormattedValue(value=BinOp(left=BinOp(left=Name(id='val', ctx=Load(), lineno=1, col_offset=3, end_lineno=1, end_col_offset=6), op=Add(), right=BinOp(left=Constant(value=1, kind=None, lineno=1, col_offset=7, end_lineno=1, end_col_offset=8), op=Mult(), right=Constant(value=34, kind=None, lineno=1, col_offset=9, end_lineno=1, end_col_offset=11), lineno=1, col_offset=7, end_lineno=1, end_col_offset=11), lineno=1, col_offset=3, end_lineno=1, end_col_offset=11), op=Add(), right=Call(func=Name(id='f', ctx=Load(), lineno=1, col_offset=12, end_lineno=1, end_col_offset=13), args=[Name(id='x', ctx=Load(), lineno=1, col_offset=12, end_lineno=1, end_col_offset=13)], keywords=[], lineno=1, col_offset=12, end_lineno=1, end_col_offset=16), lineno=1, col_offset=3, end_lineno=1, end_col_offset=16), conversion=-1, format_spec=None, lineno=1, col_offset=0, end_lineno=1, end_col_offset=18)], lineno=1, col_offset=0, end_lineno=1, end_col_offset=18), lineno=1, col_offset=0, end_lineno=1, end_col_offset=18)], type_ignores=[])
> E         + Module(body=[Expr(value=JoinedStr(values=[FormattedValue(value=BinOp(left=BinOp(left=Name(id='val', ctx=Load(), lineno=1, col_offset=3, end_lineno=1, end_col_offset=6), op=Add(), right=BinOp(left=Constant(value=1, kind=None, lineno=1, col_offset=7, end_lineno=1, end_col_offset=8), op=Mult(), right=Constant(value=34, kind=None, lineno=1, col_offset=9, end_lineno=1, end_col_offset=11), lineno=1, col_offset=7, end_lineno=1, end_col_offset=11), lineno=1, col_offset=3, end_lineno=1, end_col_offset=11), op=Add(), right=Call(func=Name(id='f', ctx=Load(), lineno=1, col_offset=12, end_lineno=1, end_col_offset=13), args=[Name(id='x', ctx=Load(), lineno=1, col_offset=14, end_lineno=1, end_col_offset=15)], keywords=[], lineno=1, col_offset=12, end_lineno=1, end_col_offset=16), lineno=1, col_offset=3, end_lineno=1, end_col_offset=16), conversion=-1, format_spec=None, lineno=1, col_offset=0, end_lineno=1, end_col_offset=18)], lineno=1, col_offset=0, end_lineno=1, end_col_offset=18), lineno=1, col_offset=0, end_lineno=1, end_col_offset=18)], type_ignores=[])
> ```

Just pushed a fix.

lysnikolaou:
@pablogsal Do you think this is enough to check the second check mark above?

pablogsal:
> Just pushed a fix.

Nice! I was planning to work in the f-string concatenation now but if you want you can start working on that as I probably won't have time until tomorrow.

pablogsal:
> @pablogsal Do you think this is enough to check the second check mark above?

Yeah, once we start adding tests, we will probably find more cases.

lysnikolaou:
@pablogsal I didn't get the time to properly work on f-string concatenation. Have you done something thus far or should I go ahead and start working on it?

pablogsal:
> Have you done something thus far or should I go ahead and start working on it?

Sadly, I am currently dealing with some health-related family stuff and I am not sure if I will be able to look at this property in the next days, so I would say to go ahead and start working on it if you have time!

lysnikolaou:
I found the time and looked into it and I think that the recent change, so that `string_token` returns a correctly parsed string, overcomplicates things when it comes to concatenating, since the `FStringParser_*` functions expect nothing more that a `char *`. I played around a bit and came up with something the works up to some expected line/col errors. In my approach, `string_token` just returns a `Token *` (much like `async_token` and such) and the `concatenate_strings` extracts them out of the `asdl_seq *` and concatenates them together. What do you think about that approach, @pablogsal, @gvanrossum?

pablogsal:
> What do you think about that approach

Hum, that approach should work and seems to allow us to re-use more of the existing machinery. The way I was going to experiment is adding some extra function to the set of functions that operate with the `FStringParser` that is able to "unpack" an existing `JoinedString` and add it to the current state of the f-string parser. Being said that I don't know if that would simplify things over the design you are proposing, which seems simpler to me.

If we change `string_token` to return a `Token*` we need to remember to adapt the rest of the grammars to such change (wrapping the result in a `Constant` or similar).

lysnikolaou:
> If we change `string_token` to return a `Token*` we need to remember to adapt the rest of the grammars to such change (wrapping the result in a `Constant` or similar).

No need to, as the one where we call `concatenate_strings` is the only occurrence of `STRING` in the grammar.

pablogsal:
> No need to, as the one where we call `concatenate_strings` is the only occurrence of `STRING` in the grammar.

I was referring to other grammars like `data/cprog.gram` (check #163) but this is probably low priority.

lysnikolaou:
> > No need to, as the one where we call `concatenate_strings` is the only occurrence of `STRING` in the grammar.
> 
> I was referring to other grammars like `data/cprog.gram` (check #163) but this is probably low priority.

Oh okay! That's right!

lysnikolaou:
Since we are not sure whether we actually adopt my approach or not, should I maybe open a new PR, which is based on this one, so that you can review what I've been doing?

pablogsal:
> Since we are not sure whether we actually adopt my approach or not, should I maybe open a new WIP PR, which is based on this one, so that you can review what I've been doing?

For me is fine if you want to keep pushing to this one. We can always revert/eliminate the new commits.

lysnikolaou:
Just pushed my work here! Here is the example shown on #188 as parsed by us and CPython:
`f"{hello} world" f"again {and_again}"`

Pegen:

```
Module(
    body=[
        Expr(
            value=JoinedStr(
                values=[
                    FormattedValue(
                        value=Name(
                            id="hello",
                            ctx=Load(),
                            lineno=1,
                            col_offset=3,
                            end_lineno=1,
                            end_col_offset=8,
                        ),
                        conversion=-1,
                        format_spec=None,
                        lineno=1,
                        col_offset=0,
                        end_lineno=1,
                        end_col_offset=37,
                    ),
                    Constant(
                        value=" worldagain ",
                        kind=None,
                        lineno=1,
                        **col_offset=17,**
                        end_lineno=1,
                        end_col_offset=37,
                    ),
                    FormattedValue(
                        value=Name(
                            id="and_again",
                            ctx=Load(),
                            lineno=1,
                            **col_offset=26,**
                            end_lineno=1,
                            **end_col_offset=35,**
                        ),
                        conversion=-1,
                        format_spec=None,
                        lineno=1,
                        **col_offset=17,**
                        end_lineno=1,
                        end_col_offset=37,
                    ),
                ],
                lineno=1,
                col_offset=0,
                end_lineno=1,
                end_col_offset=37,
            ),
            lineno=1,
            col_offset=0,
            end_lineno=1,
            end_col_offset=37,
        )
    ],
    type_ignores=[],
)
```

CPython:

```
Module(
    body=[
        Expr(
            value=JoinedStr(
                values=[
                    FormattedValue(
                        value=Name(
                            id="hello",
                            ctx=Load(),
                            lineno=1,
                            col_offset=3,
                            end_lineno=1,
                            end_col_offset=8,
                        ),
                        conversion=-1,
                        format_spec=None,
                        lineno=1,
                        col_offset=0,
                        end_lineno=1,
                        end_col_offset=37,
                    ),
                    Constant(
                        value=" worldagain ",
                        kind=None,
                        lineno=1,
                        **col_offset=0,**
                        end_lineno=1,
                        end_col_offset=37,
                    ),
                    FormattedValue(
                        value=Name(
                            id="and_again",
                            ctx=Load(),
                            lineno=1,
                            **col_offset=1,**
                            end_lineno=1,
                            **end_col_offset=10,**
                        ),
                        conversion=-1,
                        format_spec=None,
                        lineno=1,
                        **col_offset=0,**
                        end_lineno=1,
                        end_col_offset=37,
                    ),
                ],
                lineno=1,
                col_offset=0,
                end_lineno=1,
                end_col_offset=37,
            ),
            lineno=1,
            col_offset=0,
            end_lineno=1,
            end_col_offset=37,
        )
    ],
    type_ignores=[],
)
```

lysnikolaou:
I propose to change the calculated row/col for `FormattedValue` to match the CPython ones, but consider the NAME (`and_again`) row/col a CPython bug, since it's obviously wrong, and leave it as is. Maybe also open an issue on bpo for that?

gvanrossum:
I'm still struggling with a nasty cold/flu, I hope to be back to reviewing
code in a few days.

pegen/pegen.c
```diff
@@ -1677,70 +1613,1156 @@ parsestr(Parser *p, const char *s, int *bytesmode, int *rawmode, PyObject **resu
...
+    Py_XDECREF(bytes_str);
+    FstringParser_Dealloc(&state);
     return NULL;
 }
+
+void *
+string_token(Parser *p)
+{
+    return expect_token(p, STRING);
+}
```
gvanrossum: Seriously, this file is now over 2700 lines long? I worry that by the time we're done, it'll be as long as ast.c (nearly 6000). That would make arguing for this harder -- I had hoped to eliminate most of ast.c, and it looks like I vastly overestimated the possible win there... :-(

And is the f-string code in ast.c really over 1000 lines? At the very least it's probably a good idea to move it to its own file (even if that means we'll have some more public symbols).

lysnikolaou:
> Sorry folks, I don't think I can review this reasonably.

That really seems reasonable. We'll have to wait for @pablogsal then.

> If I read that correctly, we're reporting line numbers and offsets relative to the start of the string, while ast.parse() correctly reports them relative to the file?

That was indeed the case, but only for `slice_ty`s. I pushed a fix for it.

pegen/pegen.c
lysnikolaou: That's a bit sad that f-strings made things so much uglier. 1700 lines wasn't that much, but now it really is more chaotic than before. At least, I think the code is somewhat cleaner.

> At the very least it's probably a good idea to move it to its own file (even if that means we'll have some more public symbols).

Agreed!

lysnikolaou:
After the latest changes, `p->start_rule_func` gets corrupted in `fstring_compile_expr` when parsing `Lib/dataclasses.py`. I don't know what the reason is, but I'm working on it.

lysnikolaou:
> After the latest changes, `p->start_rule_func` gets corrupted in `fstring_compile_expr` when parsing `Lib/dataclasses.py`. I don't know what the reason is, but I'm working on it.

Fixed! It turns out the bug was already there!

gvanrossum:
OK, so the remaining differences with stdlib ast are all due to a bug in ast.c, right? Did you file a bug? Have you found a fix yet?

lysnikolaou:
> OK, so the remaining differences with stdlib ast are all due to a bug in ast.c, right? Did you file a bug? Have you found a fix yet?

Just created an issue. I added you in the nosy list. I didn't work on a fix, but hopefully I'll get to that later today or tomorrow.

lysnikolaou:
@gvanrossum How would go about moving the all the code about strings to a new file. I created a new header file, where I define the needed structs and the functions that need to be public and moved all of the C code to the corresponding source file. Still, I can't get it to compile, since there's a circular dependency and I'm getting multiple compilation errors.

gvanrossum:
If you make a WIP PR I can take a look. :-)

lysnikolaou:
Thanks! Opened #192.

pablogsal:
This weekend I will be able to review the new commits and improvements but on a quick inspection, they look great :)

lysnikolaou:
It seems that the stdlib "bugs" won't be solved anytime soon, since the solution isn't trivial enough to implement. What would be the correct way to proceed? Maybe mark the failing f-string tests as `xfail`?

lysnikolaou:
Also, @pablogsal, is there anything I can do here that would help things move faster?

pablogsal:
I think that having those specific tests as xfail should be the way to go.

What I think we should do is having a test case for f-strings with every representative case of an expression inside, so we know we are not missing any case.

After that we can double check the error code and merge and any polishing can be done in a new PR and this one is already gigantic. What do you think?

Btw, thanks for the help! This PR was complex and I am glad that we are working together :)

lysnikolaou:
I just had another idea. For every f-string test case we add, we could add them as xfail to the existing tests and also create a new test function for f-string that call `ast.dump(...)` without the `include_attributes=True` keyword arg. This way we can verify we generate a correct AST.

> What I think we should do is having a test case for f-strings with every representative case of an expression inside, so we know we are not missing any case.

Do you mean creating a test case for every expression type inside an f-string?

> After that we can double check the error code and merge and any polishing can be done in a new PR and this one is already gigantic. What do you think?

Sounds good!

>Btw, thanks for the help! This PR was complex and I am glad that we are working together :)

Don't mention it. Glad I could help!

pablogsal:
> I just had another idea. For every f-string test case we add, we could add them as xfail to the existing tests and also create a new test function for f-strings that calls ast.dump(...) without the include_attributes=True keyword arg. This way we can verify we generate a correct AST.

Sounds good 👍  There may be some cases that won't have problems with the node locations, so we can have them run normally.

> Do you mean creating a test case for every expression type inside an f-string?

Yes, maybe not all but the idea is to exercise the `switch` statements that we have.

> Don't mention it. Glad I could help!

:)

gvanrossum:
Ideally we would have a way to state that a test is considered passing despite a given end column mismatch. But that seems complex give the current approach (diff the ast.dump() output), so xfail seems fine, for a few tests that only verify column offsets for f-string concatenation.

Why do we need to test every representative case inside f-strings? If, say, `f"{1 + 1}"` works, I am confident that `f"{a().b()}"` also works. But we should test edge cases, e.g. `{{` and `}}`, as well as things like `!r`, `=` and `:12d`, and some edge cases (e.g. spaces before/after `=`).

pablogsal:
> Why do we need to test every representative case inside f-strings? If, say, f"{1 + 1}" works, I am confident that f"{a().b()}" also works

Yup! What I wanted to say is that I think it would be good to exercise this `switch` cases:

https://github.com/gvanrossum/pegen/pull/183/files#diff-e2161640949755a51e055a22cf4af0fdR290-R376

The idea is that we can have at least 1 example that visits every case.

gvanrossum:
Oh, given that we have that switch it makes sense to exercise every case!

pablogsal:
@lysnikolaou I have pushed some new tests. For now, it seems that lambdas and comprehensions have wrong node information in the `Name` nodes.

pablogsal:
Ok, I have fixed comprehensions, for now, only the lambda case is failing.

pablogsal:
Lambda fixed!

pablogsal:
@lysnikolaou There is something odd with this:

```
f'{(lambda: 42)(42)}'
```

The `Call` node does not have the correct col numbers, could you investigate?

lysnikolaou:
> **Edit**: Oh, actually, this may be this problem: [python/cpython#18477](https://github.com/python/cpython/pull/18477)

Yup, that's what it is!

Also two questions:
1) I'm not sure, I understand why the new `xfail` test function. It seems a bit counter intuitive to have a function that tests throwing a SyntaxError to something that does not fail (and to circumvent that mark it as `xfail`). What I'm thinking is just mark it as `xfail` and do exactly what `test_correct_ast_generation` does? But maybe I'm missing something.

2) Do you have any idea why Travis errors? It appears it cannot find `pip`, which is odd, since it never had that problem before. Does it maybe have to do with the release of 3.8.2? That doesn't seem likely though.

pablogsal:
> 1. It seems a bit counter intuitive to have a function that tests throwing a SyntaxError to something that does not fail

Whoops, that is completely wrong. As you said, it should be testing the same as `test_correct_ast_generation`. Thanks for the catch!

> Do you have any idea why Travis errors?

No idea, but it happens on master as well and I saw some people complaining in twitter recently :(

lysnikolaou:
Added more test cases with what @gvanrossum suggested. Travis now complains due to the `f'{(lambda: 3)(2)}'` test, which was solved by python/cpython#18477, but I verified locally that it's ok.

lysnikolaou:
Also, @pablogsal, is there any specific function, of which I could double-check the error code?

pablogsal:
In general, anything that was not in `ast.c` or things that were there but we modified substantially. I found a couple of places where we weren't checking for errors when calling the C-API and maybe there are a few more.

lysnikolaou:
@pablogsal, could you please review my latest commit?

pegen/parse_string.c
```diff
@@ -1195,7 +1197,6 @@ make_str_node_and_del(Parser *p, PyObject **str, Token* first_token, Token *last
         return NULL;
     }
     //TODO: Check this logic with the kind
-    //
     const char* the_str = PyBytes_AsString(first_token->bytes);
```
lysnikolaou: Can we remove this TODO now?

pegen/parse_string.c
```diff
@@ -578,9 +578,14 @@ fstring_compile_expr(Parser *p, const char *expr_start, const char *expr_end,
 
     if (asdl_seq_LEN(mod->v.Module.body) == 0) {
         raise_syntax_error(p, "f-string: empty expression not allowed");
+        goto exit;
     }
 
     stmt_ty expr = asdl_seq_GET(mod->v.Module.body, 0);
+    if (asdl_seq_LEN(mod->v.Module.body) != 1 || expr->kind != Expr_kind) {
```
pablogsal: This cannot be here as there is a check later for `mod == NULL`. If `mod` is `NULL` then `asdl_seq_LEN(mod->v.Module.body)` will segfault.

lysnikolaou: That's right, but won't `stmt_ty expr = asdl_seq_GET(mod->v.Module.body, 0);` segfault as well, then?

pablogsal: 🤦‍♂ Yep, that was wrong as well. We need to do the `NULL` check before we do anything else.

pegen/parse_string.c
pablogsal: Yep, I will do another PR after this one to make the "u" an interned string so we don't create one every time.

pegen/parse_string.c
lysnikolaou: Oh, there is a null check for `mod` on line 575. This only gets reached, in case it isn't NULL.

pablogsal: 🤦‍♂ 🤦‍♂ Note to self: do not review code in public transport 😉

lysnikolaou:
Another concern of mine is that I overlooked [this](https://github.com/python/cpython/blob/8c3aee65ed3aff3668da330ccd6f9ba7b2aa4567/Python/ast.c#L4791) part of `fstring_shift_node_locations` when rewriting it for pegen. I can't really find an example though, where this if is needed.

pablogsal:
> Edit: This could be a problem in multi line f-strings like:

We may need to pass the parent lineno down to force resets there to handle that. If this is complicated enough I would suggest to open a new issue and handle that separately.

lysnikolaou:
> If this is complicated enough I would suggest to open a new issue and handle that separately.

I think, a new issue is the way to go!

lysnikolaou:
Travis picked up the changes and the tests are all OK, so I think it's time to merge this! @pablogsal What do you say?

pablogsal:
> @pablogsal What do you say?

I say let's finally merge this 🚀 🚀

lysnikolaou:
Go for it.

PR #184, opened 2020-01-26
Some additions to Makefile

gvanrossum:
- get rid of useless `which` commands
- make simpy and simpy_cpython delete and create ./python
- add 'bench' target which runs simpy_cpython three times with error output suppressed

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -62,7 +62,7 @@ time_stdlib_compile:
 time_stdlib_parse:
 	/usr/bin/time -l $(PYTHON) -c "import ast; ast.parse(open('$(TIMEFILE)').read())"
 
-simpy: no-cpython
+simpy: clean-cpython
 	$(PYTHON) scripts/test_parse_directory.py \
 		-g data/simpy.gram \
 		-d $(TESTDIR) \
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -81,12 +81,24 @@ simpy_cpython:
 		--exclude "*/bad*" \
 		--exclude "*/lib2to3/tests/data/*"
 
+cpython:
+	tar xf data/cpython-lib.tgz
+
+no-cpython:
```
pablogsal: Nit (feel free to ignore): Maybe `clean-cpython`? The reason is that is not clear that depending on `no-cpython` will remove things.

Makefile
```diff
@@ -81,12 +81,24 @@ simpy_cpython:
 		--exclude "*/bad*" \
 		--exclude "*/lib2to3/tests/data/*"
 
+cpython:
```
pablogsal: Nit (feel free to ignore): Maybe `unpack-cpython`?

Makefile
gvanrossum: Done.

Makefile
gvanrossum: No, I want this to be the name of the cpython directory so make will extract the tarball only if needed. (But I can't make it depend on the tarball, else make will always extract it because tar sets the mtime of the extracted directory to whatever's in the tarball.)

FWIW our tarball contains a lot of extras.

PR #189, opened 2020-02-06
Add scripts to download and parse the top PyPI packages

lysnikolaou:


data/top-pypi-packages-365-days.json
```diff
@@ -0,0 +1 @@
...82,"project":"awacs"},{"download_count":1889385,"project":"azure-functions-devops-build"},{"download_count":1884376,"project":"locket"},{"download_count":1882404,"project":"ctutlz"},{"download_count":1875062,"project":"snapshottest"},{"download_count":1874184,"project":"pdfkit"},{"download_count":1870591,"project":"scapy"},{"download_count":1869037,"project":"opencensus-context"},{"download_count":1862753,"project":"flask-mail"},{"download_count":1860985,"project":"intervaltree"},{"download_count":1856012,"project":"azure-mgmt-sqlvirtualmachine"},{"download_count":1853788,"project":"azure-mgmt-kusto"},{"download_count":1853245,"project":"luigi"},{"download_count":1852083,"project":"pylru"},{"download_count":1848356,"project":"sklearn-pandas"},{"download_count":1846838,"project":"pydantic"},{"download_count":1845633,"project":"email-validator"},{"download_count":1844376,"project":"pyquery"},{"download_count":1841139,"project":"django-oauth-toolkit"},{"download_count":1839835,"project":"memory-profiler"},{"download_count":1839825,"project":"jupyterlab-server"},{"download_count":1835726,"project":"sqlalchemy-migrate"},{"download_count":1832053,"project":"retry-decorator"},{"download_c
```
gvanrossum: Why minify this? Minified JSON is good if it's part of a frequently used service. But here I'd rather be able to see what's in the list.

scripts/download_pypi_packages.py
```diff
@@ -0,0 +1,69 @@
+import os
```
gvanrossum: Maybe add a `#!` line and make the file executable? (Like test_parse_directory.py.)

scripts/download_pypi_packages.py
```diff
@@ -0,0 +1,69 @@
...
+import json
+import shutil
+
+from typing import Dict, Any
+from urllib.request import urlretrieve
+
+# This can be either the number of packages to download or 'ALL' to download
+# all 4000 provided by https://hugovk.github.io/top-pypi-packages/
+NUMBER_OF_PACKAGES = 3
```
gvanrossum: How about using argparse?

scripts/test_parse_directory.py
```diff
@@ -39,7 +39,9 @@
 argparser.add_argument(
     "--skip-actions", action="store_true", help="Suppress code emission for rule actions",
 )
-argparser.add_argument("-t", "--tree", action="count", help="Compare parse tree to official AST")
+argparser.add_argument(
+    "-t", "--tree", action="count", help="Compare parse tree to official AST", default=0
```
gvanrossum: Curious -- isn't the default 0 for `action="count"`? It always worked for me.

scripts/test_parse_directory.py
```diff
@@ -202,21 +208,29 @@ def main() -> None:
...
-    main()
+    args = argparser.parse_args()
+    directory = args.directory
+    grammar_file = args.grammar_file
+    verbose = args.verbose
+    excluded_files = args.exclude
+    skip_actions = args.skip_actions
+    tree = args.tree
+    short = args.short
+    main(directory, grammar_file, verbose, excluded_files, skip_actions, tree, short, None)
```
gvanrossum: I propose to rename main() to something else and put this code in a new argument-less main() function.

scripts/test_pypi_packages.py
```diff
@@ -0,0 +1,87 @@
+"""Usage: python -m scripts.test_pypi_packages"""
```
gvanrossum: That's a bit tricky. Maybe instead add a `#!` line, make the script executable, and add `sys.path.append('.')`?

scripts/test_pypi_packages.py
```diff
@@ -0,0 +1,87 @@
...
+            "*/lib2to3/tests/data/*",
+        ],
+        skip_actions=False,
+        tree_arg=False,
+        short=True,
+        extension=extension,
+    )
+
+
+def main() -> None:
```
gvanrossum: Would be nice if we could override some flags to test_parse_directory.py so we can emulate `-t` and `-tt`, which causes tree diffs to happen.

scripts/test_parse_directory.py
lysnikolaou: Nope! The default here is `None`, according to the docs.

scripts/test_pypi_packages.py
lysnikolaou: Is only overriding `-t` enough for now?

data/top-pypi-packages-365-days.json
lysnikolaou: I thought that it would be better to not waste memory on this. But then again, it's only around 120KBs more.

scripts/test_pypi_packages.py
gvanrossum: Sure.

scripts/test_pypi_packages.py
```diff
@@ -0,0 +1,101 @@
+#!/usr/bin/env python3.8
```
gvanrossum: chmod 0755

scripts/download_pypi_packages.py
gvanrossum: Still no `#!` line, and not executable.

scripts/download_pypi_packages.py
```diff
@@ -0,0 +1,80 @@
+import argparse
+import os
+import json
+import shutil
```
gvanrossum: This isn't actually used?

scripts/download_pypi_packages.py
```diff
@@ -0,0 +1,80 @@
...
+
+def download_package_code(name: str, package_json: Dict[Any, Any]) -> None:
+    source_index = -1
+    for idx, url_info in enumerate(package_json["urls"]):
+        if url_info["python_version"] == "source":
+            source_index = idx
+            break
+    filename = package_json["urls"][source_index]["filename"]
+    url = package_json["urls"][source_index]["url"]
+    urlretrieve(url, os.path.join("data", filename))
```
gvanrossum: Can you use a different directory than `data`? I want to be able to clean out the entire directory with a simple command, but there are also git-controlled files under `data`.

scripts/test_pypi_packages.py
lysnikolaou: Should I change something here?

gvanrossum: In the UNIX shell you should execute `chmod +x scripts/test_pypi_packages.py` and then commit the change (git detects file mode changes).

lysnikolaou: Done!

gvanrossum:
Looks like this stumbled over some Python 2 syntax (`except E, v`) in PyYAML. Not sure what to do about that. (Also, somehow it deletes the extracted files, which made diagnosing this a little harder.)

gvanrossum:
OK, here's a real AST col_offset diff in a decorator. I reduced it to this:
```
@a.b.c()
def f():
    pass
```
This produces (in part):
```
            decorator_list=[
                Call(
                    func=Attribute(
                        value=Attribute(
                            value=Name(
                                id="a",
                                ctx=Load(),
                                lineno=1,
                                col_offset=1,
                                end_lineno=1,
                                end_col_offset=2,
                            ),
                            attr="b",
                            ctx=Load(),
                            lineno=1,
                            col_offset=1,
                            end_lineno=1,
                            end_col_offset=4,
                        ),
                        attr="c",
                        ctx=Load(),
                        lineno=1,
                        col_offset=1,
                        end_lineno=1,
                        end_col_offset=6,
                    ),
                    args=[],
                    keywords=[],
                    lineno=1,
                    col_offset=1,
                    end_lineno=1,
                    end_col_offset=8,
                )
            ],
```
For `attr="c"`, the stdlib ast module produces `end_col_offset=4,` while we produce 6.

lysnikolaou:
Well, then `ast` is at fault here, right?

gvanrossum:
Hm, this actually looks like a bug in the stdlib AST. Also a slightly shorter repro uses `@a.b.c` (it doesn't have to be a call, just three names with two dots).

lysnikolaou:
> For attr="c", the stdlib ast module produces end_col_offset=4, while we produce 6.

I'm getting a somewhat different bug, where the `ast` module produces end_col_offset=6 for attr="b", while we produce 4.

For this, there is an easy enough fix, which I'm probably gonna submit tomorrow (it's gotten a bit late):

```
diff --git a/Python/ast.c b/Python/ast.c
index 9c48d71d15..bab672b295 100644
--- a/Python/ast.c
+++ b/Python/ast.c
@@ -1715,11 +1715,12 @@ ast_for_dotted_name(struct compiling *c, const node *n)
         return NULL;
 
     for (i = 2; i < NCH(n); i+=2) {
-        id = NEW_IDENTIFIER(CHILD(n, i));
+        const node *child = CHILD(n, i);
+        id = NEW_IDENTIFIER(child);
         if (!id)
             return NULL;
         e = Attribute(e, id, Load, lineno, col_offset,
-                      n->n_end_lineno, n->n_end_col_offset, c->c_arena);
+                      child->n_end_lineno, child->n_end_col_offset, c->c_arena);
         if (!e)
             return NULL;
     }
```

gvanrossum:
You're right, I misinterpreted the diff. Thanks for finding the fix for ast.c!

lysnikolaou:
I opened an [issue](https://bugs.python.org/issue39579) on bpo and also a [PR](https://github.com/python/cpython/pull/18405) for this.

PR #190, opened 2020-02-07
Test PyPI Packages: Do not remove extracted files in case of failure

lysnikolaou:


gvanrossum:
Oh wait, if the parse succeeds but the diff fails it still deletes the files. I think that's a problem with test_parse_directory.py not returning a nonzero exit status in that case though?

lysnikolaou:
> Oh wait, if the parse succeeds but the diff fails it still deletes the files. I think that's a problem with test_parse_directory.py not returning a nonzero exit status in that case though?

Right. I'm going to fix this first thing tomorrow.

lysnikolaou:
> Oh wait, if the parse succeeds but the diff fails it still deletes the files. I think that's a problem with test_parse_directory.py not returning a nonzero exit status in that case though?

Opened #191 for this.

PR #192, opened 2020-02-07
WIP: Move string handling code to new file

lysnikolaou:


lysnikolaou:
Note that this somehow compiles with clang but not gcc on my machine.

gvanrossum:
Huh, on my Mac, gcc is an alias for clang. Wild guess: maybe "strings.h" is a common filename and renaming it to something else might help?

gvanrossum:
Also, what errors do you get? That might give me a clue.

lysnikolaou:
Around 18000 lines of error messages with many error types. `gcc` compains for probably every file included in Python.h. The most common error message is something like `expected declaration specifiers or ‘...’ before PyObject`.

lysnikolaou:
> Wild guess: maybe "strings.h" is a common filename and renaming it to something else might help?

That was actually it! Thanks a lot!

lysnikolaou:
I renamed both files to `parse_string.[ch]`. Can you think of something better that this?

gvanrossum:
Excellent, those filenames are fine.

PR #193, opened 2020-02-08
Add lookahead to simple_stmt rule

lysnikolaou:
Running `simpy_cpython` became approx. 0.8 seconds faster with this. How much of a speedup can you observe?

### DIFF
--- a/data/simpy.gram
+++ b/data/simpy.gram
@@ -363,11 +363,8 @@ atom[expr_ty]:
     | &'[' list
     | &'{' dict
     | &'[' listcomp
-    | &'(' group
-    | &'(' genexp
-    | &'{' set
-    | &'{' dictcomp
-    | &'{' setcomp
+    | &'(' (group | genexp)
+    | &'{' (set | dictcomp | setcomp)
     | '...' { _Py_Constant(Py_Ellipsis, NULL, EXTRA) }
 
 list[expr_ty]:
--- a/data/simpy.gram
+++ b/data/simpy.gram
@@ -359,12 +359,9 @@ atom[expr_ty]:
     | 'None' { _Py_Constant(Py_None, NULL, EXTRA) }
     | &STRING a=STRING+ { concatenate_strings(p, a) }
     | NUMBER
-    | &'(' tuple
-    | &'[' list
-    | &'{' dict
-    | &'[' listcomp
-    | &'(' (group | genexp)
-    | &'{' (set | dictcomp | setcomp)
+    | &'(' (tuple | group | genexp)
+    | &'[' (list | listcomp)
+    | &'{' (dict | set | dictcomp | setcomp)
     | '...' { _Py_Constant(Py_Ellipsis, NULL, EXTRA) }
 
 list[expr_ty]:
--- a/data/simpy.gram
+++ b/data/simpy.gram
@@ -26,9 +26,9 @@ small_stmt[stmt_ty]:
     | &'global' global_stmt
     | &'nonlocal' nonlocal_stmt
 compound_stmt[stmt_ty]:
-    | &('@' | 'def' | ASYNC) function_def
+    | &('def' | '@' | ASYNC) function_def
     | &'if' if_stmt
-    | &('@' | 'class') class_def
+    | &('class' | '@') class_def
     | &('with' | ASYNC) with_stmt
     | &('for' | ASYNC) for_stmt
     | &'try' try_stmt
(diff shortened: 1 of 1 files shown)


gvanrossum:
I get a smaller speedup -- before, the median of 3 was 5.747 seconds, after it's 5.673 seconds. So that's only 1.3% faster.

However here's a change that moves the median down to 5.524, i.e. 4%. Together they will do 5%!

```
--- a/data/simpy.gram
+++ b/data/simpy.gram
@@ -12,17 +12,17 @@ simple_stmt[asdl_seq*]: a=';'.small_stmt+ [';'] NEWLINE { a }
 small_stmt[stmt_ty]:
     | assignment
     | e=expressions { _Py_Expr(e, EXTRA) }
-    | return_stmt
-    | import_stmt
-    | raise_stmt
-    | pass_stmt
-    | del_stmt
-    | yield_stmt
-    | assert_stmt
-    | break_stmt
-    | continue_stmt
-    | global_stmt
-    | nonlocal_stmt
+    | &'return' return_stmt
+    | &('import' | 'from') import_stmt
+    | &'raise' raise_stmt
+    | &'pass' pass_stmt
+    | &'del' del_stmt
+    | &'yield' yield_stmt
+    | &'assert' assert_stmt
+    | &'break' break_stmt
+    | &'continue' continue_stmt
+    | &'global' global_stmt
+    | &'nonlocal' nonlocal_stmt
 compound_stmt[stmt_ty]: function_def | if_stmt | class_def | with_stmt | for_stmt | try_stmt | while_stmt
 
```

gvanrossum:
I betcha you can do a similar thing to `compound_stmt` for even more gain!

gvanrossum:
Yup, try this:
```
compound_stmt[stmt_ty]:
    | &('@' | 'def' | ASYNC) function_def
    | &'if' if_stmt
    | &('@' | 'class') class_def
    | &('with' | ASYNC) with_stmt
    | &('for' | ASYNC) for_stmt
    | &'try' try_stmt
    | &'while' while_stmt
```
Also, this might be a tad quicker:
```
return_stmt[stmt_ty]: 'return' a=[expressions] { _Py_Return(a, EXTRA) }
```
Together they move the median down to 5.255, IOW ~9% faster (before you idea!).

lysnikolaou:
I pushed a commit with all your suggestions!

Also, I deleted the `pass`, `break` and `continue` rules. It shouldn't make that big of a difference, but they felt a bit overkill for just one keyword token.

lysnikolaou:
Would it be reasonable to add lookaheads to the `atom` rule as well?

Something like:

```
atom[expr_ty]:
    | NAME
    | 'True' { _Py_Constant(Py_True, NULL, EXTRA) }
    | 'False' { _Py_Constant(Py_False, NULL, EXTRA) }
    | 'None' { _Py_Constant(Py_None, NULL, EXTRA) }
    | a=STRING+ { concatenate_strings(p, a) }
    | NUMBER
    | &'(' tuple
    | &'[' list
    | &'{' dict
    | &'[' listcomp
    | &'(' group
    | &'(' genexp
    | &'{' set
    | &'{' dictcomp
    | &'{' setcomp
    | '...' { _Py_Constant(Py_Ellipsis, NULL, EXTRA) }
```

gvanrossum:
@apalala I would love your insights into optimizing the grammar!

data/simpy.gram
```diff
@@ -363,11 +363,8 @@ atom[expr_ty]:
     | &'[' list
     | &'{' dict
     | &'[' listcomp
```
gvanrossum: Why not combine these three too?

lysnikolaou: Because when we counted the number of each AST node in the stdlib we had found out that the amount of tuples, list and dicts was much greater than the amount of the atom nodes. If we were to combine these alternatives, wouldn't we lose time due to this?

pablogsal: > Because when we counted the number of each AST node in the stdlib we had found out that the amount of tuples, list and dicts was much greater

Ideally, we should repeat those stats in the future with the PyPy packages to double-check those assumptions.

gvanrossum: I don't think so, because the lookahead ensures that we take exactly one path for each start character.

data/simpy.gram
```diff
@@ -26,9 +26,9 @@ small_stmt[stmt_ty]:
     | &'global' global_stmt
     | &'nonlocal' nonlocal_stmt
 compound_stmt[stmt_ty]:
-    | &('@' | 'def' | ASYNC) function_def
+    | &('def' | '@' | ASYNC) function_def
```
gvanrossum: Clever!

data/simpy.gram
lysnikolaou: > I don't think so, because the lookahead ensures that we take exactly one path for each start character.

Of course.

lysnikolaou:
> 11% improvement for me compared to master!

Not too bad!

PR #194, opened 2020-02-08
Fix bug with decorated AsyncFunctionDef

lysnikolaou:


lysnikolaou:
> LGTM! Great catch!

Well, it was @ethanhs who caught, so credits to him!

PR #198, opened 2020-02-09
Fix bug with parenthesized targets and refactor assignment rule

### LINKED ISSUE #195: Difference in parsing assign with single element in parens
Consider the (odd) code:

```
(name) = 1
```
pegen gives:
```
Module(body=[Assign(targets=[Tuple(elts=[Name(id='name', ctx=Store())], ctx=Store())], value=Constant(value=1, kind=None), type_comment=None)], type_ignores=[])
```
CPython gives:
```
Module(body=[Assign(targets=[Name(id='name', ctx=Store())], value=Constant(value=1, kind=None), type_comment=None)], type_ignores=[])
```

### LINKED ISSUE #197: Difference with parenthized del
This one was surprisingly common.
Input:
`del(a)`
pegen:
```
Module(body=[Delete(targets=[Tuple(elts=[Name(id='a', ctx=Del())], ctx=Del())])], type_ignores=[])
```
CPython:
```
Module(body=[Delete(targets=[Name(id='a', ctx=Del())])], type_ignores=[])
```

lysnikolaou:
Fixes #195 and fixes #197.


data/simpy.gram
```diff
@@ -27,9 +27,10 @@ compound_stmt[stmt_ty]: function_def | if_stmt | class_def | with_stmt | for_stm
 
 # NOTE: yield_expression may start with 'yield'; yield_expr must start with 'yield'
 assignment:
-    | !'lambda' a=target ':' b=expression c=['=' d=yield_expression { d }] {
-         _Py_AnnAssign(construct_assign_target(p, a), b, c,
-         a->kind == Name_kind ? 1 : 0, EXTRA) }
+    | !'lambda' '(' a=ann_assign_target ')' ':' b=expression c=['=' d=yield_expression { d }] {
```
pablogsal: Question: Is there a way to join these two? Maybe something like:

`( '(' a=ann_assign_target ')' |  a=ann_assign_target )`

test/test_ast_generation.py
```diff
@@ -22,6 +22,8 @@
     ('assignment_false', 'a = False'),
     ('assignment_none', 'a = None'),
     ('assignment_true', 'a = True'),
+    ('assignment_paren', '(a) = 42'),
```
pablogsal: I would add also some correct examples with annotations:

```
(x) : int
(x): int = 34
```

data/simpy.gram
lysnikolaou: I don't think so, since the action is different. When there is a parenthesis, the `simple` argument to `_Py_AnnAssign` must be 1.

test/test_ast_generation.py
lysnikolaou: But there are such examples already. Should I add more?

lysnikolaou:
I think that we can now dump the `!'lambda'` lookahead, since we handle keywords correctly. Is that correct?

data/simpy.gram
```diff
@@ -27,9 +27,10 @@ compound_stmt[stmt_ty]: function_def | if_stmt | class_def | with_stmt | for_stm
 
 # NOTE: yield_expression may start with 'yield'; yield_expr must start with 'yield'
 assignment:
-    | !'lambda' a=target ':' b=expression c=['=' d=yield_expression { d }] {
-         _Py_AnnAssign(construct_assign_target(p, a), b, c,
-         a->kind == Name_kind ? 1 : 0, EXTRA) }
+    | '(' a=ann_assign_target ')' ':' b=expression c=['=' d=yield_expression { d }] {
+        _Py_AnnAssign(a, b, c, 0, EXTRA)}
+    | a=ann_assign_target ':' b=expression c=['=' d=yield_expression { d }] {
+        _Py_AnnAssign(a, b, c, 1, EXTRA) }
```
gvanrossum: If you can do `a: int` and `(a): int` then you can also do `((a)): int` etc. (All of which IMO is stupid and only accepted due to a bug in the grammar but that's backwards compatibility for you.)

test/test_ast_generation.py
pablogsal: Oh, I may have missed then. If they already exists, ignore what I said

test/test_ast_generation.py
```diff
@@ -444,6 +454,10 @@ def f(*a, b):
 ]
 
 FAIL_TEST_CASES = [
+    ("annotation_multiple_targets", "(a, b): int = 42"),
+    ("annotation_nested_tuple", "((a, b)): int"),
```
gvanrossum: I'd ad `(a,): int` and `a,: int`

data/simpy.gram
```diff
@@ -438,18 +439,25 @@ star_target[expr_ty]:
...
 del_targets[asdl_seq*]: a=','.del_target+ [','] { a }
 del_target[expr_ty]:
     | a=t_primary '.' b=NAME !t_lookahead { _Py_Attribute(a, b->v.Name.id, Del, EXTRA) }
     | a=t_primary b=slicing !t_lookahead { _Py_Subscript(a, b, Del, EXTRA) }
     | del_t_atom
 del_t_atom[expr_ty]:
     | a=NAME { set_expr_context(p, a, Del) }
-    | '(' b=[del_targets] ')' { _Py_Tuple(b, Del, EXTRA) }
-    | '[' b=[del_targets] ']' { _Py_List(b, Del, EXTRA) }
+    | '(' a=[del_target !','] ')' { set_expr_context(p, a, Del) }
```
gvanrossum: I don't see why you need the `!','` here, since you already have `')'` following.

data/simpy.gram
```diff
@@ -469,5 +477,6 @@ t_primary[expr_ty]:
 t_lookahead: '(' | '[' | '.'
 t_atom[expr_ty]:
     | a=NAME { set_expr_context(p, a, Store) }
+    | '(' a=[target !','] ')' { set_expr_context(p, a, Store) }
```
gvanrossum: Ditto.

data/simpy.gram
```diff
@@ -439,10 +439,16 @@ star_target[expr_ty]:
...
+    | '(' a=[star_target] ')' { set_expr_context(p, a, Store) }
     | '(' a=[star_targets_seq] ')' { _Py_Tuple(a, Store, EXTRA) }
     | '[' a=[star_targets_seq] ']' { _Py_List(a, Store, EXTRA) }
 
+paren_ann_assign_target[expr_ty]:
+    | '(' a=paren_ann_assign_target ')' { a }
+    | a=t_primary '.' b=NAME !t_lookahead { _Py_Attribute(a, b->v.Name.id, Store, EXTRA) }
+    | a=t_primary b=slicing !t_lookahead { _Py_Subscript(a, b, Store, EXTRA) }
+    | a=NAME { set_expr_context(p, a, Store) }
+
```
gvanrossum: Couldn't this just be
```
  | '(' a=paren_ann_assign_target ')' { a }
  | ann_assign_target
```
?

data/simpy.gram
```diff
@@ -27,7 +27,7 @@ compound_stmt[stmt_ty]: function_def | if_stmt | class_def | with_stmt | for_stm
 
 # NOTE: yield_expression may start with 'yield'; yield_expr must start with 'yield'
 assignment:
-    | '(' a=ann_assign_target ')' ':' b=expression c=['=' d=yield_expression { d }] {
+    | '(' a=paren_ann_assign_target ')' ':' b=expression c=['=' d=yield_expression { d }] {
         _Py_AnnAssign(a, b, c, 0, EXTRA)}
     | a=ann_assign_target ':' b=expression c=['=' d=yield_expression { d }] {
         _Py_AnnAssign(a, b, c, 1, EXTRA) }
```
gvanrossum: These seem to be redundant except for the `0` or `1` argument to `_Py_AnnAssign()` -- and why do we need that? (I can't even figure out when simple=1 is ever set in an AST made by `ast.parse()`?)

lysnikolaou: `simple=1` when there aren't any parentheses, `simple=0` otherwise.

In Python/ast.c there is the following snippet:

```
int simple = 1;
...
if (NCH(deep) > 0 && TYPE(CHILD(deep, 0)) == LPAR) {
    simple = 0;
}
```

gvanrossum: Hm, I tried this:
```
>>> ast.dump(ast.parse('(a):int'))
"Module(body=[AnnAssign(target=Name(id='a', ctx=Store()), annotation=Name(id='int', ctx=Load()), value=None, simple=0)], type_ignores=[])"
>>>
```
And this still comes up with `simple=0` (no matter how many parens I put around it).

lysnikolaou: Right. And that's because there are parentheses around the NAME, hence it's not simple (`simple=0`). But if you do:

```
>>> t = 'a: int'
>>> ast.dump(ast.parse(t))
"Module(body=[AnnAssign(target=Name(id='a', ctx=Store()), annotation=Name(id='int', ctx=Load()), value=None, simple=1)], type_ignores=[])"
```

you get `simple=0`.

gvanrossum: D'oh. I had the sense of the flag backwards. :-(

PR #199, opened 2020-02-09
Correctly parse number literals

### LINKED ISSUE #196: Incorrectly parses imaginary literals
For `1j` pegen parses it as `1+0j`

pablogsal:
Closes #196

### DIFF
--- a/pegen/pegen.c
+++ b/pegen/pegen.c
@@ -373,6 +373,82 @@ dedent_token(Parser *p)
     return expect_token(p, DEDENT);
 }
 
+
+static PyObject *
+parsenumber_raw(const char *s)
+{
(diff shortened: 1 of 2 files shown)


pegen/pegen.c
```diff
@@ -373,50 +373,107 @@ dedent_token(Parser *p)
     return expect_token(p, DEDENT);
 }
 
+
```
gvanrossum: These two functions are unchanged from ast.c right?

pablogsal: Yeah, except for some arguments that we don't need now (the compiling structure pointer).

gvanrossum:
OK land it.

PR #200, opened 2020-02-09
Avoid false SyntaxError's in test_parse_directory 

lysnikolaou:
Don't report invalid syntax errors, when ast.parse cannot parse file either.

When testing PyPI this happens a lot due to Python 2 code.

### DIFF
--- a/scripts/test_parse_directory.py
+++ b/scripts/test_parse_directory.py
@@ -182,10 +182,15 @@ def parse_directory(
                 if not short:
                     report_status(succeeded=True, file=file, verbose=verbose)
             except Exception as error:
-                report_status(
-                    succeeded=False, file=file, verbose=verbose, error=error, short=short
-                )
-                errors += 1
+                try:
+                    ast.parse(file)
+                except Exception:
+                    pass
+                else:
+                    report_status(
+                        succeeded=False, file=file, verbose=verbose, error=error, short=short
+                    )
+                    errors += 1
             files.append(file)
     t1 = time.time()
 
--- a/scripts/test_parse_directory.py
+++ b/scripts/test_parse_directory.py
@@ -185,7 +185,8 @@ def parse_directory(
                 try:
                     ast.parse(file)
                 except Exception:
-                    pass
+                    if not short:
+                        print(f"File {file} cannot be parsed by either pegen or the ast module.")
                 else:
                     report_status(
                         succeeded=False, file=file, verbose=verbose, error=error, short=short


scripts/test_parse_directory.py
```diff
@@ -182,10 +182,15 @@ def parse_directory(
...
                     report_status(succeeded=True, file=file, verbose=verbose)
             except Exception as error:
-                report_status(
-                    succeeded=False, file=file, verbose=verbose, error=error, short=short
-                )
-                errors += 1
+                try:
+                    ast.parse(file)
+                except Exception:
+                    pass
```
pablogsal: Nit: Can we log something here so we know that those files fail also in Python 3?

lysnikolaou:
Are we confident enough that the two won't fail for a different reason though?

pablogsal:
> Are we confident enough that the two won't fail for a different reason though?

Well, `ast.parse` may fail if they are using `async` and the code is not compatible with the latest CPython in which `async` is a soft keyword.

lysnikolaou:
> > Are we confident enough that the two won't fail for a different reason though?
> 
> Well, `ast.parse` may fail if they are using `async` and the code is not compatible with the latest CPython in which `async` is a soft keyword.

Were there syntax changes since `async` was introduced altogether?

pablogsal:
> Have there been syntax changes since `async` was introduced altogether?

I don't think so. You can use `ast.parse(feature_version=(major, minor))` to parse "old" versions.

lysnikolaou:
> > Are we confident enough that the two won't fail for a different reason though?
> 
> Well, `ast.parse` may fail if they are using `async` and the code is not compatible with the latest CPython in which `async` is a soft keyword and similar situations. But I doubt that that is common with the most popular Pypi packages.

I'm not sure I can see how this could happen. Could you give mme an example of such a code snippet?

pablogsal:
> I'm not sure I can see how this could happen. Could you give mme an example of such a code snippet?

In Python3.6 this is legal for example:

```
my_function(async=True)

or

async = 32
``

lysnikolaou:
> > I'm not sure I can see how this could happen. Could you give mme an example of such a code snippet?
> 
> In Python3.6 this is legal for example:
> 
> ```python
> my_function(async=True, await=False)
> ```
> 
> or
> 
> ```python
> async = await = 32
> ```
> 
> But those will fail in Python3.7 and later (https://docs.python.org/3/whatsnew/3.7.html#summary-release-highlights). This means that a simple `ast.parse` failing does not necessarily mean Python2 code, it may be due to one of these situations for instance.

But pegen fails for these as well, so isn't it okay that `ast.parse` fails?

pablogsal:
> But pegen fails for these as well, so isn't it okay that `ast.parse` fails?

Yup, both will fail. I was just clarifying that `ast.parse` may fail due to other reasons than Python 2 code (or purely invalid code).

lysnikolaou:
> > Are we confident enough that the two won't fail for a different reason though?

All clear now! Thanks for the explanation!

PR #206, opened 2020-02-17
Set simple=0 on attribute and subscript assignments

isidentical:


pablogsal:
I think this may be more complex than it needs to be as I suspect this is enough to fix this:

```
diff --git a/data/simpy.gram b/data/simpy.gram
index 567199c..9ca9e5e 100644
--- a/data/simpy.gram
+++ b/data/simpy.gram
@@ -39,7 +39,7 @@ assignment:
     | '(' a=paren_ann_assign_target ')' ':' b=expression c=['=' d=yield_expression { d }] {
         _Py_AnnAssign(a, b, c, 0, EXTRA)}
     | a=ann_assign_target ':' b=expression c=['=' d=yield_expression { d }] {
-        _Py_AnnAssign(a, b, c, 1, EXTRA) }
+        _Py_AnnAssign(a, b, c, a->kind == Name_kind ? 1 : 0, EXTRA) }
     | a=(z=star_targets '=' { z })+ b=(yield_expr | expressions) {
          _Py_Assign(a, b, NULL, EXTRA) }
     | a=target b=augassign c=(yield_expr | expressions) {
```

On the other hand, I also find having conditionals in actions a bit ugly. What do you think about this @gvanrossum @lysnikolaou ?

lysnikolaou:
> I think this may be more complex than it needs to be as I suspect this is enough to fix this (also having extra rules may affect performance):
> 
> ```diff
> diff --git a/data/simpy.gram b/data/simpy.gram
> index 567199c..9ca9e5e 100644
> --- a/data/simpy.gram
> +++ b/data/simpy.gram
> @@ -39,7 +39,7 @@ assignment:
>      | '(' a=paren_ann_assign_target ')' ':' b=expression c=['=' d=yield_expression { d }] {
>          _Py_AnnAssign(a, b, c, 0, EXTRA)}
>      | a=ann_assign_target ':' b=expression c=['=' d=yield_expression { d }] {
> -        _Py_AnnAssign(a, b, c, 1, EXTRA) }
> +        _Py_AnnAssign(a, b, c, a->kind == Name_kind ? 1 : 0, EXTRA) }
>      | a=(z=star_targets '=' { z })+ b=(yield_expr | expressions) {
>           _Py_Assign(a, b, NULL, EXTRA) }
>      | a=target b=augassign c=(yield_expr | expressions) {
> ```
> 
> On the other hand, I also find having conditionals in actions a bit ugly. What do you think about this @gvanrossum @lysnikolaou ?

I agree, that conditionals in actions are ugly and we should implement logic in the grammar rules themselves wherever possible. Still, I would refactor some of the rules (for example the ann_assign_rule isn't needed anymore). I would do it this way:

```
diff --git a/data/simpy.gram b/data/simpy.gram
index b84a618..0272fd0 100644
--- a/data/simpy.gram
+++ b/data/simpy.gram
@@ -36,12 +36,11 @@ compound_stmt[stmt_ty]:

 assignment:
-    | '(' a=paren_ann_assign_target ')' ':' b=expression c=['=' d=yield_expression { d }] {
+    | a=('(' b=paren_ann_assign_target ')' { b }
+         | non_simple_ann_assign_target) ':' b=expression c=['=' d=yield_expression { d }] {
         _Py_AnnAssign(a, b, c, 0, EXTRA)}
-    | a=non_simple_ann_assign_target ':' b=expression c=['=' d=yield_expression { d }] {
-        _Py_AnnAssign(a, b, c, 0, EXTRA) }
-    | a=ann_assign_target ':' b=expression c=['=' d=yield_expression { d }] {
-        _Py_AnnAssign(a, b, c, 1, EXTRA) }
+    | a=NAME ':' b=expression c=['=' d=yield_expression { d }] {
+        _Py_AnnAssign(set_expr_context(p, a, Store), b, c, 1, EXTRA) }
     | a=(z=star_targets '=' { z })+ b=(yield_expr | expressions) {
          _Py_Assign(a, b, NULL, EXTRA) }
     | a=target b=augassign c=(yield_expr | expressions) {
@@ -444,14 +443,12 @@ star_atom[expr_ty]:
 paren_ann_assign_target[expr_ty]:
     | '(' a=paren_ann_assign_target ')' { a }
     | non_simple_ann_assign_target
-    | ann_assign_target
+    | a=NAME { set_expr_context(p, a, Store) }

 non_simple_ann_assign_target[expr_ty]:
     | a=t_primary '.' b=NAME !t_lookahead { _Py_Attribute(a, b->v.Name.id, Store, EXTRA) }
     | a=t_primary b=slicing !t_lookahead { _Py_Subscript(a, b, Store, EXTRA) }

-ann_assign_target[expr_ty]: a=NAME { set_expr_context(p, a, Store) }
-
 del_targets[asdl_seq*]: a=','.del_target+ [','] { a }
 del_target[expr_ty]:
     | a=t_primary '.' b=NAME !t_lookahead { _Py_Attribute(a, b->v.Name.id, Del, EXTRA) }
```

pablogsal:
I would suggest also doing a simple check of the performance impact of adding new rules vs the (ugly) conditional so we are sure that we will not introduce performance regressions.

lysnikolaou:
> I would suggest also doing a simple check of the performance impact of adding new rules vs the (ugly) conditional so we are sure that we will not introduce performance regressions.

Good point. Also, I guess that `simple=1` assignments are much more frequent, so reordering the rules to have those appear first would probably improve performance, although I'm not sure by how much.

gvanrossum:
I've nothing to add, somebody please do the perf research and propose a final patch (maybe as a separate PR).

isidentical:
I did a performance test with directly timeit, looks like it isn't sensitive as I thought. So I'll give it a try with pyperf tomorrow.

lysnikolaou:
I did a small performance evaluation with `make simpy_cpython` and it got from 6,370 seconds (mean) on master to 6,434, but my results are highly inconsistent, so my system might not be the best to get the evaluation from. For example, yesterday I was getting 6,059 seconds on master and the only change since then is the unimportant #210.

pablogsal:
For micro-optimizations you need to activate CPU isolation a fix the CPU frequency (CPUs nowadays change frequency dynamically) as the system jitter may be too high for very subtle differences.Check out [this](https://vstinner.github.io/journey-to-stable-benchmark-system.html) for more information. For the speed.python.org server we use pyperf with all the system tuned (check [this](https://pyperf.readthedocs.io/en/latest/system.html) for context).

lysnikolaou:
After I followed the instruction outlined in the article on my significantly slower Linux machine, I get 12,062 seconds for master and  11,932 seconds for this branch.

pablogsal:
I think I like the current state of the PR and it seems that is faster due to the reordering/inlining so I think is good to merge :)

data/simpy.gram
```diff
@@ -441,12 +442,12 @@ star_atom[expr_ty]:
 
 paren_ann_assign_target[expr_ty]:
     | '(' a=paren_ann_assign_target ')' { a }
-    | ann_assign_target
+    | non_simple_ann_assign_target
+    | a=NAME { set_expr_context(p, a, Store) }
 
-ann_assign_target[expr_ty]:
+non_simple_ann_assign_target[expr_ty]:
```
pablogsal: I find a bit weird that this is called `non_simple_ann_assign_target` but there is not a `simple_ann_assign_target`. I know where the `non_simple` cames from, but I think is leaking the abstraction slightly.

data/simpy.gram
```diff
@@ -441,12 +442,12 @@ star_atom[expr_ty]:
 
 paren_ann_assign_target[expr_ty]:
```
pablogsal: Suggestion: rename this to `inside_paren_ann_assign_target` now that we are at it

data/simpy.gram
```diff
@@ -441,12 +442,12 @@ star_atom[expr_ty]:
...
     | '(' a=paren_ann_assign_target ')' { a }
-    | ann_assign_target
+    | non_simple_ann_assign_target
+    | a=NAME { set_expr_context(p, a, Store) }
```
pablogsal: Shouldn't these two be moved upwards as they would be more common?

data/simpy.gram
lysnikolaou: Maybe something like `ann_assign_target_subscript_attribute`?

lysnikolaou:
Thanks a lot, @isidentical!

gvanrossum:
Huh, looks like several other PRs ended up in this?

lysnikolaou:
This was me pulling the remote branch and merging with master. I checked the master commit and it only contains the original changes.

PR #207, opened 2020-02-17
Correctly shift node position in multi-line fstrings

### LINKED ISSUE #203: Correctly shift expr col_offset for multi-line f-strings
In a multi-line f-string the col_offset needs to be reset, in case the child node is on a different line than the parent node.

lysnikolaou:
I mark this as WIP, because I'm not confident enough that this covers all cases. Any ideas for more tests?

Fixes #203.

### DIFF
--- a/pegen/parse_string.c
+++ b/pegen/parse_string.c
@@ -258,14 +258,14 @@ static void fstring_shift_expr_locations(expr_ty n, int lineno, int col_offset);
 static void fstring_shift_argument(expr_ty parent, arg_ty args, int lineno, int col_offset);
 
 
-static inline void SHIFT_EXPR(expr_ty parent, expr_ty n, int line, int col) {
+static inline void shift_expr(expr_ty parent, expr_ty n, int line, int col) {
     if (parent->lineno < n->lineno) {
         col = 0;
     }
     fstring_shift_expr_locations(n, line, col);
 }
 
-static inline void SHIFT_ARG(expr_ty parent, arg_ty n, int line, int col) {
+static inline void shift_arg(expr_ty parent, arg_ty n, int line, int col) {
     if (parent->lineno < n->lineno) {
         col = 0;
     }
@@ -278,7 +278,7 @@ static void fstring_shift_seq_locations(expr_ty parent, asdl_seq *seq, int linen
         if (expr == NULL){
             continue;
         }
-        SHIFT_EXPR(parent, expr, lineno, col_offset);
+        shift_expr(parent, expr, lineno, col_offset);
     }
 }
 
@@ -286,13 +286,13 @@ static void fstring_shift_slice_locations(expr_ty parent, slice_ty slice, int li
     switch (slice->kind) {
         case Slice_kind:
             if (slice->v.Slice.lower) {
-                SHIFT_EXPR(parent, slice->v.Slice.lower, lineno, col_offset);
+                shift_expr(parent, slice->v.Slice.lower, lineno, col_offset);
             }
             if (slice->v.Slice.upper) {
-                SHIFT_EXPR(parent, slice->v.Slice.upper, lineno, col_offset);
+                shift_expr(parent, slice->v.Slice.upper, lineno, col_offset);
             }
             if (slice->v.Slice.step) {
-                SHIFT_EXPR(parent, slice->v.Slice.step, lineno, col_offset);
+                shift_expr(parent, slice->v.Slice.step, lineno, col_offset);
             }
             break;
         case ExtSlice_kind:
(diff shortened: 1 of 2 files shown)


gvanrossum:
I hope Pablo can review this... I'm not at all confident I understand the f-string code.

There's one thing that concerns me though -- why focus all attention on the column offset? Isn't the line number also subject to movement? E.g. in
```
f"""blah
{
expr)
}
"""
```
Don't we want the error to point to the `)` rather than to the start of the f-string?

lysnikolaou:
> I hope Pablo can review this... I'm not at all confident I understand the f-string code.
> 
> There's one thing that concerns me though -- why focus all attention on the column offset? Isn't the line number also subject to movement? E.g. in
> 
> ```
> f"""blah
> {
> expr)
> }
> """
> ```
> 
> Don't we want the error to point to the `)` rather than to the start of the f-string?

That's pretty much impossible at the moment. All the shifting around only happens after the complete AST tree has been generated. Thus, if the parsing fails, these function don't even get called.

I think, that our best shot at getting this 100% right is rewriting the whole thing, which would be quite a big project. In my imagination, we could have a grammar for f-strings and in `concatenate_strings` we would just concatenate the string literals together and then call the f-string parser on the big concatenated string if needed. In case we decide this is important, I could slowly start trying things out in that direction.

Thoughts, @gvanrossum @pablogsal?

lysnikolaou:
> why focus all attention on the column offset? Isn't the line number also subject to movement?

Because it seems that line number movement is ok, at least in comparison to what CPython does.

pablogsal:
> That's pretty much impossible at the moment. All the shifting around only happens after the complete AST tree has been generated. Thus, if the parsing fails, these function don't even get called.

Check out the WIP in: https://github.com/gvanrossum/pegen/pull/208. I have tried to provide better error messages for f-strings there. Even if we don't merge it in its entirety, that PR fixes some errors when parsing strings and not files (the filename is `<string>`) that have syntax errors.

pablogsal:
I quickly checked the PR and the `SHIFT_EXPR` and `SHIFT_ARG` look fine under a preliminary inspection but I need to certainly dedicate more time to understand what's going on here:

https://github.com/gvanrossum/pegen/pull/207/files#diff-e2161640949755a51e055a22cf4af0fdR489-R505

Could you maybe add some comments there for context meanwhile?

gvanrossum:
I think we should strive for parity with CPython first.

I agree that in the long run we may have to rewrite f-string parsing from scratch -- and Eric V Smith seems to agree (see https://bugs.python.org/issue39564), but that's going to require changes to the tokenizer. If this gets done in 3.9 we'll have to track that, but my money is on 3.10 or later, given the complexity of the edge cases (things like `...{x:4d}...` and `...{x = :.5g}...` -- and note that currently the walrus doesn't work, e.g. `...{x := 1}...`).

pablogsal:
> but that's going to require changes to the tokenizer

Good point. I think it will also likely require us to maintain a stack of tokenizers (so you can alternate between parsing f-strings and "regular Python").

lysnikolaou:
> I quickly checked the PR and the `SHIFT_EXPR` and `SHIFT_ARG` looks 👍 under a preliminary inspection but I need to certainly dedicate more time to understand what's going on here:
> 
> https://github.com/gvanrossum/pegen/pull/207/files#diff-e2161640949755a51e055a22cf4af0fdR489-R505
> 
> Could you maybe add some comments there for context meanwhile?

Done.

pegen/parse_string.c
```diff
@@ -472,13 +486,27 @@ fstring_fix_expr_location(Token *parent, expr_ty n, char *expr_str)
...
-            start = substr;
-            while (start > parent_str) {
-                if (start[0] == '\n')
+            // The following is needed, in order to correctly shift the column
+            // offset, in the case that (disregarding any whitespace) a newline
+            // immediately follows the opening curly brace of the fstring expression.
+            int newline_after_brace = 1;
+            start = substr + 1;
+            while (start && *start != '}' && *start != '\n') {
+                if (!(*start == ' ' || *start == '\t' || *start == '\f')) {
```
pablogsal: Suggestion, introduce the "!" inside the condition to simplify the logic (so it reads directly left to right without the need to negate everything at the end) :

```
(*start != ' ' && *start != '\t' && *start != '\f')
```

pegen/parse_string.c
```diff
@@ -472,13 +486,27 @@ fstring_fix_expr_location(Token *parent, expr_ty n, char *expr_str)
...
+                    newline_after_brace = 0;
                     break;
-                start--;
+                }
+                start++;
+            }
+            if (!newline_after_brace) {
+                start = substr;
+                while (start > parent_str) {
```
pablogsal: Can you just use `(start > parent_str && start[0] != '\n')` as the `while` condition and remove the `if` and the `break`?

pegen/parse_string.c
```diff
@@ -472,13 +486,27 @@ fstring_fix_expr_location(Token *parent, expr_ty n, char *expr_str)
...
+            start = substr + 1;
+            while (start && *start != '}' && *start != '\n') {
+                if (!(*start == ' ' || *start == '\t' || *start == '\f')) {
+                    newline_after_brace = 0;
                     break;
-                start--;
+                }
+                start++;
+            }
+            if (!newline_after_brace) {
```
pablogsal: Suggestion: Maybe add a comment similar to "Account for the characters from the last newline character to our left until the beginning of `substr`"

pegen/parse_string.c
```diff
@@ -445,7 +457,9 @@ static void fstring_shift_children_locations(expr_ty n, int lineno, int col_offs
 static void fstring_shift_expr_locations(expr_ty n, int lineno, int col_offset)
 {
     n->col_offset = n->col_offset + col_offset;
-    n->end_col_offset = n->end_col_offset + col_offset;
+    if (n->lineno == n->end_lineno) {
```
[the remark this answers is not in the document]
lysnikolaou: No, this is not the only case, where this is true. For example, in the fstring

```
f'''
{
a
}
'''
```

the condition `n->lineno == n->end_lineno` holds, while in the fstring

```
f'''
{
f(
    a, b, c
)
}
'''
```

it is not true.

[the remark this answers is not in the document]
pablogsal: I think that it would help to add a comment explaining roughtly what situations are we correcting here because is not inmediately obvious, but this is a "nice to have" so if you find it not straightfoward to describe is fine :)

pegen/parse_string.c
```diff
@@ -255,185 +255,199 @@ parsestr(Parser *p, const char *s, int *bytesmode, int *rawmode, PyObject **resu
 // FSTRING STUFF
 
 static void fstring_shift_expr_locations(expr_ty n, int lineno, int col_offset);
+static void fstring_shift_argument(expr_ty parent, arg_ty args, int lineno, int col_offset);
 
-static void fstring_shift_seq_locations(asdl_seq *seq, int lineno, int col_offset) {
+
+static inline void SHIFT_EXPR(expr_ty parent, expr_ty n, int line, int col) {
```
pablogsal: Nit: Now that they are functions these should be lowercase i think.

lysnikolaou:
Thanks a lot for the review @pablogsal!

PR #208, opened 2020-02-18
Provide better tracebacks with f-strings

pablogsal:
With this PR:

```
venv ❯ python -c 'from pegen import parse; parse.parse_string("f\"{x$4+34+34}\"")'
  File "<string>", line 1
    (x$4+34+34)
      ^
SyntaxError: invalid syntax
```

### DIFF
--- a/pegen/pegen.c
+++ b/pegen/pegen.c
@@ -43,7 +43,6 @@ raise_syntax_error(Parser *p, const char *errmsg, ...)
     PyObject *errstr = NULL;
     PyObject *loc = NULL;
     PyObject *tmp = NULL;
-    PyObject *filename = NULL;
     Token *t = p->tokens[p->fill - 1];
     va_list va;
 
@@ -53,20 +52,16 @@ raise_syntax_error(Parser *p, const char *errmsg, ...)
     if (!errstr) {
         goto error;
     }
-    if (p->tok->filename && PyUnicode_CompareWithASCIIString(p->tok->filename, "<string>") != 0) {
+    if (p->input_mode == 0) {
         if (PyErr_Occurred()){
             goto error;
         }
-        filename = p->tok->filename;
-        loc = PyErr_ProgramTextObject(filename, t->lineno);
+        loc = PyErr_ProgramTextObject(p->tok->filename, t->lineno);
         if (!loc) {
             Py_INCREF(Py_None);
             loc = Py_None;
         }
-    }
-    else {
-        Py_INCREF(Py_None);
-        filename = Py_None;
+    } else {
         loc = PyUnicode_FromString(p->tok->buf);
         if (!loc) {
             goto error;
@@ -77,7 +72,7 @@ raise_syntax_error(Parser *p, const char *errmsg, ...)
     // situations involving invalid indentation.
     int col_offset = t->col_offset == -1 ? 0 : t->col_offset;
     Py_ssize_t col_number = byte_offset_to_character_offset(loc, col_offset) + 1;
-    tmp = Py_BuildValue("(OiiN)", filename, t->lineno, col_number, loc);
+    tmp = Py_BuildValue("(OiiN)", p->tok->filename, t->lineno, col_number, loc);
     if (!tmp) {
         goto error;
     }
@@ -93,9 +88,6 @@ raise_syntax_error(Parser *p, const char *errmsg, ...)
 
 error:
     Py_XDECREF(errstr);
-    if (!p->tok->filename) {
-        Py_XDECREF(filename);
-    }
     Py_XDECREF(loc);
     return -1;
 }
@@ -507,7 +499,7 @@ keyword_token(Parser *p, const char *val)
 
 PyObject *
 run_parser(struct tok_state *tok, void *(start_rule_func)(Parser *), int mode,
-           KeywordToken **keywords, int n_keyword_lists)
+           int input_mode, KeywordToken **keywords, int n_keyword_lists)
 {
     PyObject *result = NULL;
     Parser *p = PyMem_Malloc(sizeof(Parser));
@@ -517,6 +509,7 @@ run_parser(struct tok_state *tok, void *(start_rule_func)(Parser *), int mode,
     }
     assert(tok != NULL);
     p->tok = tok;
+    p->input_mode = input_mode;
     p->keywords = keywords;
     p->n_keyword_lists = n_keyword_lists;
     p->tokens = PyMem_Malloc(sizeof(Token *));
@@ -558,7 +551,7 @@ run_parser(struct tok_state *tok, void *(start_rule_func)(Parser *), int mode,
     }
 
     if (mode == 2) {
-        result = (PyObject *)PyAST_CompileObject(res, tok->filename, NULL, -1, p->arena);
+        result = (PyObject *) PyAST_CompileObject(res, tok->filename, NULL, -1, p->arena);
     }
     else if (mode == 1) {
         result = PyAST_mod2obj(res);
@@ -609,7 +602,7 @@ run_parser_from_file(const char *filename, void *(start_rule_func)(Parser *), in
     tok->filename = filename_ob;
     filename_ob = NULL;
 
-    result = run_parser(tok, start_rule_func, mode, keywords, n_keyword_lists);
+    result = run_parser(tok, start_rule_func, mode, 0, keywords, n_keyword_lists);
 
     PyTokenizer_Free(tok);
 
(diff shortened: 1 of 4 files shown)


pegen/pegen.c
```diff
@@ -53,7 +53,10 @@ raise_syntax_error(Parser *p, const char *errmsg, ...)
     if (!errstr) {
         goto error;
     }
-    if (p->tok->filename) {
+    if (p->tok->filename && PyUnicode_CompareWithASCIIString(p->tok->filename, "<string>") != 0) {
```
pablogsal: I would like to avoid to call `PyUnicode_CompareWithASCIIString` and instead make the other branch work better.

lysnikolaou:
I kind of have mixed feelings about this. I really really like being able to point to the exact location of the fstring that's failing. OTOH, I really dislike `File "<string>", line 1`. In a bigger file, it might be difficult to locate the f-string that's failing, without knowing which line it's on.

pablogsal:
> I really dislike `File "<string>", line 1`

That is because i am using `python -c`. If you use a file:

```
venv ❯ python the_file.py                                                         
  File "the_file.py", line 3, in <module>
    parse.parse_string("f\"{x{})+34+34}\"")
  File "<string>", line 1
    f"{x{})+34+34}"
                  ^
SyntaxError: f-string: unmatched ')'
```

pablogsal:
> I really dislike `File "<string>", line 1`

@lysnikolaou That is because I am using `python -c`. If you use a file:

```
venv ❯ python the_file.py                                                         
  File "the_file.py", line 3, in <module>
    parse.parse_string("f\"{x{})+34+34}\"")
  File "<string>", line 1
    f"{x{})+34+34}"
                  ^
SyntaxError: f-string: unmatched ')'
```

lysnikolaou:
Dumb question coming: If we have something like
```
foo = 1
bar = 2
baz = f'{x$y}'
```
where would pegen integrated into CPython point to?

pablogsal:
> Edit: Well, any decent editor provides some kind of search functionality, which might make it a bit easier to locate the error, but do we want to rely on users searching for the fstring text?

Consider this file:

```
def f():
  ...
  y = f"{x$34}"
  ...

x = 42
```

this is what **CPython** produces:

```
venv ❯ python other_file.py
  File "<fstring>", line 1
    (x$34)
      ^
SyntaxError: invalid syntax
```

this is what pegen produces:

```
venv ❯ python -c 'from pegen import parse; parse.parse_file("other_file.py")'
  File "<string>", line 1
    (x$34)
      ^
SyntaxError: invalid syntax
```

pablogsal:
> where would pegen integrated into CPython point to in the error message and what line number would be displayed? That's my concern.

Unless I am missing something, with this PR both will point to the same place unless the f-string is multi-line (because we need to cover that case in this PR). With your example:

```
~/github/pegen better_fstrings*
venv ❯ cat other_file.py
foo = 1
bar = 2
baz = f'{x$y}'

~/github/pegen better_fstrings*
venv ❯ python other_file.py                                                  
  File "<fstring>", line 1
    (x$y)
      ^
SyntaxError: invalid syntax

~/github/pegen better_fstrings*
venv ❯ python -c 'from pegen import parse; parse.parse_file("other_file.py")'
  File "<string>", line 1
    (x$y)
      ^
SyntaxError: invalid syntax

```

lysnikolaou:
All clear now! I was under the illusion that CPython outputs the correct line number. Thanks for the thorough explanation.

pegen/pegen.c
lysnikolaou: That'd be good, because we could rename the filename to `fstring` in `fstring_compile_expr`, like CPython does, more comfortably.

lysnikolaou:
I think, that if we can fix the `if` mentioned in the review and improve the code in `raise_syntax_error`, we can land this. I'm not confident enough to do this though.

pablogsal:
I plan to finish it as soon as possible (and review the other PR) but I am still dealing with the family health issues that I mentioned last time so it may take a day or two more until I have time to sit down and do it. Hopefully, I will have more availability in a week or so :)

If you want to go ahead you could try to work on the 'if' branch. What I think we need to do there is using the token information to grab the correct line and column for the fstring. The other branch is already doing that with the `PyErr_ProgramTextObject` but that reads a file first and does a bunch of extra stuff while what we need to do here is use the buffer directly.

lysnikolaou:
> I plan to finish it as soon as possible (and review the other PR) but I am still dealing with the family health issues that I mentioned last time so it may take a day or two more until I have time to sit down and do it. Hopefully, I will have more availability in a week or so :)

No worries. I hope everything is well.

> If you want to go ahead you could try to work on the 'if' branch. What I think we need to do there is using the token information to grab the correct line and column for the fstring. The other branch is already doing that with the PyErr_ProgramTextObject but that reads a file first and does a bunch of extra stuff while what we need to do here is use the buffer directly.

I'll try my best!

lysnikolaou:
@pablogsal I looked into this and have found out two thing that are interesting:

1) I can't think of any other way to delete the `PyUnicode_CompareWithASCIIString` call in the `if` branch other than having `p->tok->filename` be NULL (there is no other way to know if the input comes from a file or not) and then set it in `raise_syntax_error` to something like `<string>` for calls to `parse_string` and `<fstring>` for expressions within f-strings. That means that we would probably need another parameter `is_in_fstring`, in order to distinguish between the two.

2) The very simple call `PyUnicode_FromString(p->tok->buf)` appears to be working correctly, because `p->tok->buf` only contains the input, *starting from* the error line. Then only the first line of the buffer gets printed in the traceback, which means that this is okay. 

In case there are no errors in what I've found out, I could certainly go ahead and implement this.

pablogsal:
> * I can't think of any other way to delete the `PyUnicode_CompareWithASCIIString` call in the `if` branch other than having `p->tok->filename` be NULL (there is no other way to know if the input comes from a file or not) and then set it in `raise_syntax_error` to something like `<string>` for calls to `parse_string` and `<fstring>` for expressions within f-strings. That means that we would probably need another parameter `is_in_fstring`, in order to distinguish between the two.

We could add a new attribute to the `Parser` struct called `mode` or something like that means `file` or `string`. The whole design of `raise_syntax_error` *may* change substantially when we integrate it into CPython, so I would not worry too much for now. Is also better to have the definitive design decisions at that moment as we may be missing some design considerations at the moment.

> Then only the first line of the buffer gets printed in the traceback, which means that this is okay.

Oh, I may be missing something because when I was playing around that way I was getting some weird tracebacks but if we add some tests to double-check that the tracebacks look nice I don't see any downside to that approach.

> In case there are no errors in what I've found out, I could certainly go ahead and implement this.

👍  Let's add some tests to check the tracebacks to be sure that we don't forget about these cases if we refactor and, especially, in the CPython integration.

lysnikolaou:
Should I push to this PR or open a new one?

pablogsal:
> Should I push to this PR or open a new one?

Feel free to push to this one if you want :)

pegen/pegen.c
```diff
@@ -609,7 +602,7 @@ run_parser_from_file(const char *filename, void *(start_rule_func)(Parser *), in
     tok->filename = filename_ob;
     filename_ob = NULL;
 
-    result = run_parser(tok, start_rule_func, mode, keywords, n_keyword_lists);
+    result = run_parser(tok, start_rule_func, mode, 0, keywords, n_keyword_lists);
```
pablogsal: Suggestion: Let's use an enum for the 1 and 0 values so we know what they mean when reading the code

pegen/pegen.h
```diff
@@ -7,6 +7,12 @@
 #include <Python-ast.h>
 #include <pyarena.h>
 
+enum INPUT_MODE {
```
pablogsal: If you do `typedef enum  {...} INPUT_MODE;` you can use INPUT_MODE as a type :
```
    INPUT_MODEinput_mode;  // Where the input comes from (0 for file, 1 for string, 2 for fstring)
```

lysnikolaou: Oh, that's what you meant in your other comment. On it!

pegen/pegen.h
```diff
@@ -7,6 +7,12 @@
...
+    FILE_INPUT,
+    STRING_INPUT,
+    FSTRING_INPUT
```
pablogsal: Do we need to distinguish `FSTRING_INPUT` from `STRING_INPUT`? The filename string will be different but the behaviour is the same, no?

lysnikolaou: Right, I just thought that this could be somehow useful in the future. Should I remove it?

pablogsal: > Right, I just thought that this could be somehow useful in the future. Should I remove it?

As you prefer, I would say yes just to make it a bit more clear that there are only two modes right now to understand.

lysnikolaou: You know best.

pablogsal:
I think this is a great improvement (thanks a lot for the help and the great work so we can speed fixing all these issues). I would merge it if you are ok with it :)

lysnikolaou:
Go ahead!

pablogsal:
🚀

PR #212, opened 2020-02-21
Avoid reallocating memory too often for children asdl_seq in loop functions

lysnikolaou:
Are you getting a speedup from this? I don't really trust my Linux for benchmarking.


lysnikolaou:
I'm getting a 7% speedup from this.

PR #213, opened 2020-02-21
Reorder alternatives in expression rule

lysnikolaou:
I'm getting a speedup of almost 17% from this.


lysnikolaou:
> Makes sense. (Can’t try it right now so I take your word for it.)

Never mind. It was due to the bug mentioned in #214 and most files getting parsed up to a certain point and then failing. For some reason, I wasn't getting the failure locally.

PR #214, opened 2020-02-21
Reject invalid expressions in the for-if clauses of a comprehension

lysnikolaou:
For example, pegen currently accepts the invalid `(a for a in lambda: b)`.

I found this out, while investigating the Travis failures in #213. This also stresses the need for testing with invalid input.

### DIFF
--- a/data/simpy.gram
+++ b/data/simpy.gram
@@ -388,7 +388,7 @@ kvpair[KeyValuePair*]:
     | '**' a=bitwise_or { key_value_pair(p, NULL, a) }
     | a=expression ':' b=expression { key_value_pair(p, a, b) }
 for_if_clauses[asdl_seq*]:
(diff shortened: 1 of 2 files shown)


gvanrossum:
+1 on the test fix too.

lysnikolaou:
> How did you find this?

I was investigating the Travis failures from #213, which all were related to this.

> Also, eventually this may point to some possible relaxation of the grammar? Or would there be some ambiguity that PEG resolves by taking the first match?

What do you mean relaxation of the grammar?

gvanrossum:
Relaxation: turn e.g. disjunction into expression. But in this case probably there’s a reason.

lysnikolaou:
> Relaxation: turn e.g. disjunction into expression.

But doesn't this PR then do the opposite of relaxation?

gvanrossum:
Indeed, but presumably that to match the old parser.

I was hypothesizing that using pegen will allow us to see some restrictions in the grammar and let us relax them. Though on second thought probably not this case.

PR #216, opened 2020-02-21
Only store lineno/col_offset if EXTRA is used in the action

lysnikolaou:
Not a significant speedup, but this avoids going into the loop in `get_last_non_whitespace_token` if it's not needed.

### DIFF
--- a/pegen/c_generator.py
+++ b/pegen/c_generator.py
@@ -366,10 +366,8 @@ class CParserGenerator(ParserGenerator, GrammarVisitor):
                 with self.indent():
                     self.print("return res;")
             self.print("int mark = p->mark;")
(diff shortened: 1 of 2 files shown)


pegen/c_generator.py
```diff
@@ -366,7 +366,10 @@ def _handle_default_rule_body(self, node: Rule, rhs: Rhs, result_type: str) -> N
                 with self.indent():
                     self.print("return res;")
             self.print("int mark = p->mark;")
-            self._set_up_token_start_metadata_extraction()
+            for alt in rhs.alts:
```
gvanrossum: Please code this using ‘any()’ instead of loop and break.

PR #217, opened 2020-02-21
Measure true parsing speed, in case the produced AST tree is not needed

lysnikolaou:


gvanrossum:
Hm, this still measures butecode compilation time IIUC

lysnikolaou:
> Hm, this still measures butecode compilation time IIUC

Right. I pushed a fix.

PR #222, opened 2020-02-26
Make the empty tuple a valid target

lysnikolaou:
While investigating the `simpy_cpython` failure on #130, I found out that statements with the empty tuple as target such as `del()`, `for () in a: pass` and `with a as (): pass` get rejected.

### DIFF
--- a/data/simpy.gram
+++ b/data/simpy.gram
@@ -435,7 +435,7 @@ star_target[expr_ty]:
     | star_atom
 star_atom[expr_ty]:
     | a=NAME { set_expr_context(p, a, Store) }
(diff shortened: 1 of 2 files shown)


gvanrossum:
I'm confused. Before this PR, `del ()` seemed to work too.

pablogsal:
The problem for  is that the rule was allowing to call the action:

```
set_expr_context(p, a, Store)
```

with `a=NULL`

lysnikolaou:
> Before this PR, del () seemed to work too.

That's odd. It wasn't working for me.

PR #225, opened 2020-03-01
Add CHECK_CALL to grammar and improve error checking code

### LINKED ISSUE #224: Improve error checking code in grammar helpers

lysnikolaou:
Closes #224.


pegen/pegen.c
```diff
@@ -767,15 +756,13 @@ seq_count_dots(asdl_seq *seq)
     int number_of_dots = 0;
     for (int i = 0, l = asdl_seq_LEN(seq); i < l; i++) {
         Token *current_expr = asdl_seq_GET(seq, i);
+        assert(current_expr->type == ELLIPSIS || current_expr->type == DOT);
         if (current_expr->type == ELLIPSIS) {
             number_of_dots += 3;
         }
         else if (current_expr->type == DOT) {
```
pablogsal: If we remove the `else` then the `if` is not needed in this line. We could also transform this into a switch for clarity (and maybe some speed).

pegen/pegen.c
```diff
@@ -796,31 +783,13 @@ alias_for_star(Parser *p)
     return alias(str, NULL, p->arena);
 }
 
-/* Returns the first element of seq or previous if seq is empty */
-void *
-seq_get_head(void *previous, asdl_seq *seq)
```
pablogsal: We should remove this two from the header file and the `helpers.md` file now that they are not used anymore.

pegen/pegen.c
```diff
@@ -875,24 +849,15 @@ _get_exprs(Parser *p, asdl_seq *seq)
     return new_seq;
 }
 
-/* Wrapper for _Py_Compare, so that the call in the grammar stays concise */
-expr_ty
-Pegen_Compare(Parser *p, expr_ty expr, asdl_seq *pairs, int lineno, int col_offset,
```
pablogsal: We should remove this from the header file and the helpers.md file now that it's not used anymore.

pegen/pegen.h
```diff
@@ -103,13 +102,21 @@ void *dedent_token(Parser *p);
...
 #define UNUSED(expr) do { (void)(expr); } while (0)
 #define EXTRA_EXPR(head, tail) head->lineno, head->col_offset, tail->end_lineno, tail->end_col_offset, p->arena
 #define EXTRA start_lineno, start_col_offset, end_lineno, end_col_offset, p->arena
 
+inline void *
+CHECK_CALL(Parser *p, void *result)
+{
+    if (result == NULL) {
+        longjmp(p->error_env, 1);
+    }
```
pablogsal:
```suggestion
    if (result == NULL) {
        assert(PyErr_Occurred());
        longjmp(p->error_env, 1);
    }
```

lysnikolaou: Should this be an assertion or should it be in the if condition? This way, we could use `CHECK_CALL` for *every* nested call. Some functions are allowed and should be able to return NULL, like get_keys and get_values in `dictrule`.

pablogsal: >  Some functions are allowed and should be able to return NULL, like get_keys and get_values in dictrule.

Those return always an `asdl_seq *` that is freshly created, no?

```
asdl_seq *
get_keys(Parser *p, asdl_seq *seq)
{
    int len = asdl_seq_LEN(seq);
    asdl_seq *new_seq = _Py_asdl_seq_new(len, p->arena);
    if (!new_seq) {
        return NULL;
    }
    for (int i = 0; i < len; i++) {
        KeyValuePair *pair = asdl_seq_GET(seq, i);
        asdl_seq_SET(new_seq, i, pair->key);
    }
    return new_seq;
}
```

What can be NULL there?

It should be an assertion because we don't need to differentiate between `NULL`-as-error and some other meaning of `NULL`. If a helper returns `NULL` is always an error. The assert will allow us to detect cases when we are not setting the error indicator or something like that.

pegen/pegen.c
```diff
@@ -757,11 +757,13 @@ seq_count_dots(asdl_seq *seq)
     for (int i = 0, l = asdl_seq_LEN(seq); i < l; i++) {
         Token *current_expr = asdl_seq_GET(seq, i);
         assert(current_expr->type == ELLIPSIS || current_expr->type == DOT);
```
pablogsal: Suggestion: We can move this assert to a new "default" case. It may help if some compiler complains that we are not using all values in the `switch`.

pegen/pegen.h
lysnikolaou: That's right. I was getting some errors in `simpy_cpython` that were related to the empty `dict`. I had initially included the `CHECK_CALL` call for `get_keys` and `get_values` and the errors went away, when I removed those. At the time, I assumed that `_Py_asdl_seq_new` returns NULL, if called with `size=0`, but that's not the case.

I now added them back and everything runs smoothly, so it must have been something else, but I don't know exactly what.

In any case, my bad.

lysnikolaou:
> If a helper returns NULL is always an error. 

There are also the functions `seq_{extract|delete}_starred_exprs`, which return NULL and that shouldn't be an error. What are we doing with those?

pablogsal:
> What are we doing with those?

Can we factor out the parts that return `NULL` for other reasons to the call site (if needed)? Maybe we need to enclose these functions in some helper that checks for error only on those or use another macro that also checks for `PyErr_Occured()`.

lysnikolaou:
@pablogsal Would something like [`8cd74ed`](https://github.com/gvanrossum/pegen/pull/225/commits/8cd74ed702d310577ae6ecd0d64e5a4021a76f6c) work?

Also, could you quickly test out if there's an unexpected performance regression on your system?

pablogsal:
> Also, could you quickly test out if there's an unexpected performance regression on your system?

On it

pablogsal:
> On it

Nothing that I can detect. Microbenchmarks for individual rules show some micro-seconds of difference for the extra check but that is expected. I checked what happens when compiled with PGO with `perf` (https://perf.wiki.kernel.org/index.php/Main_Page) and seems that the happy path branch is highly predictable so no problem.

lysnikolaou:
> > On it
> 
> Nothing that I can detect. Microbenchmarks for individual rules show some micro-seconds of difference for the extra check but that is expected. I checked what happens when compiled with PGO with `perf` (https://perf.wiki.kernel.org/index.php/Main_Page) and seems that the happy path branch is highly predictable so no problem.

Nice.

Thanks a lot for the review!

gvanrossum:
Sorry, I realize it's a bit late, since this has landed already. But for readability, I kind of think that the macro could reference `p` without it being passed in (since `EXTRA` does the same), and since it's pretty common, it could be named just `CHECK()`.

lysnikolaou:
> Sorry, I realize it's a bit late, since this has landed already. But for readability, I kind of think that the macro could reference `p` without it being passed in (since `EXTRA` does the same), and since it's pretty common, it could be named just `CHECK()`.

It's never too late. Opened #227.

PR #232, opened 2020-03-20
Use list instead of set in collect_todo

lysnikolaou:
This way the parser gets generated deterministically and can be included to version control, which is needed when integrating pegen to CPython.


pablogsal (approved):

pegen/parser_generator.py
```diff
@@ -73,10 +73,10 @@ def printblock(self, lines: str) -> None:
             self.print(line)
 
     def collect_todo(self) -> None:
-        done: Set[str] = set()
+        done: List[str] = list()
         while True:
-            alltodo = set(self.todo)
-            todo = alltodo - done
+            alltodo = list(self.todo)
```
pablogsal: Hummm, `alltodo` can still remain a set, no?

pegen/parser_generator.py
```diff
@@ -73,10 +73,10 @@ def printblock(self, lines: str) -> None:
             self.print(line)
 
     def collect_todo(self) -> None:
-        done: Set[str] = set()
+        done: List[str] = list()
```
pablogsal: Ideally this should remain a `set` to not make the `not in done` part quadratic when considering the loop.

lysnikolaou: But if it's a set and insertion order is not guaranteed, rule generation order is not deterministic, which is our goal here, right? What about both of them being dicts and working with their keys, would that work?

pablogsal: You are not iterating over `done` so the insertion order does not matter. Instead of `done = alltodo` you can do `done = set(alltodo)` no?

lysnikolaou: Oh, you mean that only `alltodo` becomes a list and `done` remains a set. That seems reasonable.

lysnikolaou:
Do you think that adding parse.c to version control is a good idea, since its generation is not random anymore?

pegen/parser_generator.py
pablogsal: Yeah, what I was referring is that the `not in done` part is `O(n)` if `done` is a list.

pablogsal:
> Do you think that adding parse.c to version control is a good idea, since its generation is not random anymore?

I would be +1

gvanrossum (approved):

gvanrossum:
Thanks! I take it that a bit,of,this now needs to be down-ported to we-like-parsers?

lysnikolaou:
> Thanks! I take it that a bit of this now needs to be down-ported to we-like-parsers?

Yeah, I'm on it.

PR #233, opened 2020-03-23
Allow expressions in yield_expression

lysnikolaou:
Prior to this, assignments like `x: tuple = 1, 2` were not accepted.


gvanrossum:
Hm, isn't that intentional?

lysnikolaou:
> Hm, isn't that intentional?

But `x: tuple = 1, 2` is valid in 3.8.

pablogsal:
I think this was done here:

https://github.com/python/cpython/pull/13760

Issue #220, opened 2020-02-24
Improve memory usage


pablogsal:
In order to get the performance of pegen closer to the original parser, we need to profile and improve memory usage. This can be achieved now that we have 100% AST coverage (modulo some small tweaks with column and line numbers) so we don't expect big refactorings that will introduce object creation.

- [ ] Benchmark the current pegen master against the CPython parser.
- [ ] Identify/research a method to pinpoint what creates more memory to get improvement opportunities.
- [ ] Iterate until we get as close as we can to the original parser memory usage.

gvanrossum:
I suspect that one of the most useful techniques will be to selectively disable the memoization cache for certain situations. At least, some early feedback I received when I blogged about PEG was that the memoization is wasteful (it is of course also essential in some cases, so we'll have to be careful).

What kind of memory profiling tools do we have?

pablogsal:
> What kind of memory profiling tools do we have?

In the past, I have used Massif (https://valgrind.org/docs/manual/ms-manual.htmt) but I certainly would need to re-learn how to use it as I only used it once or twice.

ethanhs:
I have used https://pypi.org/project/memory-profiler/ before, but that may not exactly lend itself to C code.

gvanrossum:
I did a quick experiment, disabling memoization for all rules (except left-recursive ones, which need it). This reduced the memory usage quite a bit, but also slowed parsing down. But it confirms my hunch that we need to tone down the memoization. My next experiment will be to instrument the memoization to see which rules, if any, are ever used from the memo.

gvanrossum:
OK, so the vast majority of the cache hits are due to left-recursive rules (the top one being `term`), and we can't touch those, since the left-recursion implementation depends on the memoization cache.

The first non-left-recursive cache hit is `atom`, with an order of magnitude fewer hits than `term`. I'm going to change my instrumentation to measure the size of the thing being cached (probably just the number of tokens, since that's the only number I have easily available).

pablogsal:
> OK, so the vast majority of the cache hits are due to left-recursive rules (the top one being `term`), and we can't touch those, since the left-recursion implementation depends on the memoization cache.

What workload are you using to test? Parsing the standard library?

gvanrossum:
Just one file for now -- Lib/tkinter/__init__.py.

gvanrossum:
[...]

PR #236, opened 2020-03-26
Add memoization control (for now, C only)

gvanrossum:
Addresses https://github.com/gvanrossum/pegen/issues/220#issuecomment-604019170 

Memoization is expensive. In the C generator, we now disable it except for rules with the special marker `(memo)` after the rule name (and type, if present). By selectively turning on memoization for a handful of rules the parser becomes faster and uses less memory.

Notes:
- Left-recursive rules always use memoization, since the implementation of left-recursion depends on it.
- The marker syntax is extensible, in case we need more per-rule flags in the future.

This also adds instrumentation to the C support code to measure how much each rule uses memoization.

Finally, it fixes two lines in scripts/test_parse_directory.py that were still using pegen/ instead of peg_parser/ as the output directory.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -49,7 +49,7 @@ stats: peg_parser/parse.c
 time: time_compile
 
 time_compile: peg_parser/parse.c
-	/usr/bin/time -l $(PYTHON) -c "from peg_parser import parse; parse.parse_file('$(TIMEFILE)', mode=2); parse.dump()"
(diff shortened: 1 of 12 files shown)


gvanrossum:
This will require precarious porting to the cpython branch at https://github.com/we-like-parsers/cpython :-(

gvanrossum:
> I think I'd prefer to not have whitespace between the type and the flags.

Hm, I added that intentionally so the flags stand out a bit more -- it's a bit busy already with the type in square brackets.

I'll just land, so we can find out the consequences for other work soon.

PR #237, opened 2020-03-27
Use the arena to allocate individual tokens

gvanrossum:
This gives a modest improvement in space usage.

### DIFF
--- a/peg_parser/parse_string.c
+++ b/peg_parser/parse_string.c
@@ -601,7 +601,7 @@ fstring_compile_expr(Parser *p, const char *expr_start, const char *expr_end,
         PyErr_Format(PyExc_MemoryError, "Out of memory for tokens");
         goto exit;
     }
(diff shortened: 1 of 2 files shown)


gvanrossum:
This uses 1% less memory than master to compile data/xxl.txt, and compiles it 5% faster.

Unfortunately it uses 1.5% *more* memory to compile `tkinter/__init__.py` (that file is much smaller and the difference in compilation speed is too small to measure).

pablogsal:
> Unfortunately it uses 1.5% _more_ memory to compile `tkinter/__init__.py` (that file is much smaller and the difference in compilation speed is too small to measure).

Is this because the arena block overhead?

gvanrossum:
Hm, thinking more about it, this is wrong -- it extends the lifetime of the array of tokens until the arena is freed, which happens only after bytecode compilation is complete. Presumably the small gain for data/xxl.txt is because that file is 100,000 times the same constant expression.

I have a few other ideas though.

PR #238, opened 2020-03-27
Don't pre-allocate all Token structs

gvanrossum:
Timing results:

- data/xxl.txt: 5% less memory, 1% faster
- tkinter/__init__.py: a tiny bit more memory (within the noise)

Also get rid of unused token_name() function completely.

PS. it's unfortunate that the Parser allocation/free code is duplicated in parse_string.c.

### DIFF
--- a/peg_parser/parse_string.c
+++ b/peg_parser/parse_string.c
@@ -596,13 +596,11 @@ fstring_compile_expr(Parser *p, const char *expr_start, const char *expr_end,
     p2->input_mode = STRING_INPUT;
     p2->keywords = p->keywords;
     p2->n_keyword_lists = p->n_keyword_lists;
-    p2->tokens = PyMem_Malloc(sizeof(Token *));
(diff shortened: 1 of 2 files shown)


gvanrossum:
Still no convincing timings. For the stdlib it's all a wash (the variability is much more than the difference between branches).

lysnikolaou:
> PS. it's unfortunate that the Parser allocation/free code is duplicated in parse_string.c.

Oops, I fixed that on the cpython repo, but didn't downport it here. 

Would we be interested in a bot that automatically opens PRs when a file exists in both repos? I guess, we could fork miss-islington and work with it or create something from scratch. It shouldn't be too hard to implement, but it still seems like a distraction.

gvanrossum:
I think such a bot would be more work than occasionally moving changes back and forth. But if you could port this particular change that would be nice. I will try to find time to port my memo-related speedup to the cpython repo.

lysnikolaou:
> But if you could port this particular change that would be nice.

Of course, I'll do it a bit later today.

PR #251, opened 2020-10-19
Add Github Actions and remove Travis

lysnikolaou:


lysnikolaou:
I guess, we now need to fix `pytest` and get rid of the Travis webhook.

gvanrossum (approved):
Thanks! Landing...

PR #252, opened 2020-10-19
Don't skip over test_ast_generation and test_c_parser

lysnikolaou:


gvanrossum (commented):
Can we rename "test" to "tests" so the "from test.support import ..." things actually work without setting CPYTHON_ROOT=...?

pegen/build.py
```diff
@@ -64,10 +64,14 @@ def compile_c_extension(
...
         cpython_root: pathlib.Path = pathlib.Path(cpython_root_str)
     else:
-        cpython_root = MOD_DIR.parent.parent.parent
+        cpython_root = MOD_DIR.parent.parent / "cpython"
         if not (cpython_root / "Python").is_dir():
             # This is Guido's convention. :-)
             cpython_root = pathlib.Path.home() / "cpython"
+        if not (cpython_root / "Python").is_dir():
+            raise ValueError(
+                "No CPython repository found. Please use the CPYTHON_ROOT env varaible."
```
gvanrossum, line 73:
```suggestion
                "No CPython repository found. Please use the CPYTHON_ROOT env variable."
```

gvanrossum (approved):
But fix typo.

pablogsal (approved):

